Issue #1, opened 2017-06-24
Add video release date to list views


stevenvolckaert:
At the moment, the release date of videos is not visible in list views. Adding the release date would make it much easier to identify the most recent video.

stevenvolckaert:
Closed by merging #6.

dagwieers:
We added the video release data also to the metadata, so every video entry now shows this information at the right bottom (in my skin).

PR #6, opened 2017-06-25
Added video broadcast date to video items

stevenvolckaert:
Fixes issue #1.

Prepended the title of each video item (Kodi list items) with the broadcast date (release date).

This makes videos much easier to identify, as it sometimes takes several hours before a new episode comes available online.

At the moment, I'm using the date format as it is displayed in the tag with the `tile__broadcastdate--mobile` class. As a further improvement, we can (should?) parse this date using [`datetime.strptime()`][1], and format it in another, less ambiguous format (e.g. "25 JUN 2017" or "2017-06-25" for today).

Hope you like my contribution!

[1]: https://docs.python.org/2/library/datetime.html#strftime-and-strptime-behavior

### DIFF
--- a/resources/lib/vrtplayer/metadatacollector.py
+++ b/resources/lib/vrtplayer/metadatacollector.py
@@ -13,7 +13,7 @@ class MetadataCollector:
         metadata_creator = metadatacreator.MetadataCreator()
         metadata_creator.duration = self.__get_episode_duration(soup)
         metadata_creator.plot = self.get_plot(soup)
-        metadata_creator.date = self.get_broadcast_date(soup)
+        metadata_creator.datetime = self.get_broadcast_datetime(soup)
         return metadata_creator.get_video_dictionary()
 
     def get_multiple_layout_episode_metadata(self, soup):
@@ -22,7 +22,7 @@ class MetadataCollector:
         return metadata_creator.get_video_dictionary()
 
     @staticmethod
-    def __get_episode_duration( soup):
+    def __get_episode_duration(soup):
         duration = None
         duration_item = soup.find(class_="content__duration")
         if duration_item is not None:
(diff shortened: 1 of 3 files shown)


stevenvolckaert:
I encountered an error when loading the details of a category if that category only contains a single item. It probably got introduced by my changes in the `MetadataCollector.get_broadcast_date(soup)` method, as that one is called (I think) when there's only one video item.

So, please wait from merging this request. I'll investigate this weekend. If you would already have an idea what's going wrong, drop a comment here. Greets!

resources/lib/vrtplayer/metadatacollector.py
```diff
@@ -50,6 +52,22 @@ def get_plot(soup):
         return description
 
     @staticmethod
+    def get_broadcast_date(soup):
+        broadcast_date_tag = soup.find(class_="content__broadcastdate")
+        
+        if broadcast_date_tag is None:
```
pietje666: Could you rewrite this so the method has only one return statement. I do not like multiple return statements in methods.   When debugging i often set a breakpoint on the bottom return statement and then it would not hit. Thanks

pietje666:
> I encountered an error when loading the details of a category if that category only contains a single item. It probably got introduced by my changes in the 

Indeed 'get_single_layout_episode_metadata'  gets called only for videos with a single episode available. 
So right now the date would only show only on videos with a single episode with your current changes (correct me if i am wrong). I think this is a case where the date is less important then when there are multiple episodes for one video.

About appending the date to the title, is adding to the metadata not enough or does it not show it visually on first sight?

stevenvolckaert:
In the current state of this branch, the date is prepended to the title field **if there is more than one episode**. That's done in `VRTPlayer.get_single_video(self, path, soup)` and `VRTPlayer.__get_item(element, is_playable)`.

If there's only a single episode, it seems like this logic isn't executed, but the `MetadataCollector.get_broadcast_date(soup)` instead. It is there that the error occurs (will investigate this weekend).

I'm setting the metadata (the `date` property) in the expected Kodi format (`%d.%m.%Y`), but this doesn't seem render the date in the list views. I'm using the [Eminence][1] skin, though.

I'll investigate further this weekend with the default skin.

[1]: http://xbmc-skins.com/skin/eminence

resources/lib/vrtplayer/metadatacollector.py
stevenvolckaert: Will do - I was following the [*fail fast*][1] principle, as explained in an [article by Martin Fowler][2] (quite an interesting read, by the way - would recommend it). I'm following this principle for years, it's a great help in my experience.

I'll adapt this weekend.

[1]: https://en.wikipedia.org/wiki/Fail-fast
[2]: https://martinfowler.com/ieeeSoftware/failFast.pdf

stevenvolckaert:
The bug is fixed; see my last commit. This branch is now ready to be merged.

I checked with Kodi's default skin, [Estuary][1]: Like with [Eminence][2], when creating a list item, the `date` value from the metadata dictionary is *not* displayed in the list. So, it seems that prepending the date to every list item name is required if we want to make it visible.

Have an awesome day!

[1]: http://kodi.wiki/view/Add-on:Estuary
[2]: http://xbmc-skins.com/skin/eminence

pietje666:
Looks good. A bit annoying the date is not being shown out off the box by Kodi (it does use the date for sorting i think).

PR #20, opened 2018-05-07
Fix stream urls

yorickps:
VRT is now using a CDN (vualto / akamai) and is using new stream links (https://services.vrt.be/videoplayer/r/live.json). The old links still work, however when there is lots of traffic / viewers Kodi has problems keeping up with the stream. The CDN links should fix that.


pietje666:
Thanks for pull request! Fyi, due to a holiday i will not be able to merge this before the end off the month.

PR #39, opened 2018-11-18
support for drm livestreams

mediaminister:
Hello,

I added support for the DRM-protected 720p livestreams. You can switch DRM on in the addon settings.
You need the latest Kodi 18 and inputstream.adaptive with libwidevinecdm.

### DIFF
--- a/plugin.video.vrt.nu/resources/language/resource.language.en_gb/strings.po
+++ b/plugin.video.vrt.nu/resources/language/resource.language.en_gb/strings.po
@@ -32,6 +32,14 @@ msgctxt "#32011"
 msgid "Show subtitles if available"
 msgstr "Show subtitles if available"
 
+msgctxt "#32020"
+msgid "DRM"
+msgstr "DRM"
+
+msgctxt "#32021"
+msgid "Use DRM for 720p HD-quality livestreams"
+msgstr "Use DRM for 720p HD-quality livestreams"
+
 msgctxt "#32051"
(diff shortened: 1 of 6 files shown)


pietje666:
perfect was planning to do the same in the future, will merge later this week when i have time to look into it with more detail

dagwieers:
Nice, installing widevine is easiest by using Kodi18 (LibreELEC 8.90.007+), enabling the VideoPlayer InputStream addon named "InputStream Adaptive" and installing widevine using the Retrospect addon settings (which performs every step for you). Worked like a charm on Raspberry Pi.

Can't wait to test this, do we have any known DRM streams to test this with ?

dagwieers:
Would be nice if we could have the same logic implemented as Retrospect has, or at least add instructions to the Wiki, to make this as painless as possible.

dagwieers:
Yes, in Retrospect settings the user can click through (only when the right dependencies have been installed) to install widevine. On Raspberry Pi it will download a 1.9GB image, and then extracts the library and moves it in the right location. And it reports success.

https://bitbucket.org/basrieter/xbmc-online-tv/src/f62a3d22afe83bdbddc36e67369e95bde53c6ac8/net.rieter.xot/resources/menus/menu_install_widevine.py?at=master&fileviewer=file-view-default

I assume that on other platforms it takes a more simple approach, as the 1.9GB download is a Raspberry Pi workaround for licensing issues.

PR #44, opened 2018-12-19
faster stream selection

mediaminister:
I edited UrlToStreamService class to provide the stream urls faster. For compatibility with older Kodi installations without inputstream.adaptive or DRM-support, the highest quality HLS-variant is directly provided cause the standard Kodi-player is fairly slow to select the HLS-variants.

So, inputstream.adaptive or DRM isn't necessarily required.

### DIFF
--- a/plugin.video.vrt.nu/resources/lib/vrtplayer/urltostreamservice.py
+++ b/plugin.video.vrt.nu/resources/lib/vrtplayer/urltostreamservice.py
@@ -1,5 +1,5 @@
 import requests
-import urlparse
+from urlparse import urljoin
 import datetime
 import time
 import _strptime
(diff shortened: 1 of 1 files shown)


pietje666:
pushed a new version to the kodi repo with all your changes. Btw how do you test if your code is working right now? + Do you have a background in developing mediastreams or something?

dagwieers:
All working fine. 720p on live streams, w00t w00t !
And subtitles is working fine as well.

Is it possible to get higher resolution content for on-demand watching too ?

pietje666:
So right now you create a Zip, manually install it in Kodi for testing? There are also Unit test classes available which might make your life easier when testing :).

mediaminister:
No, I created a symbolic link between the Kodi addon folder and the plugin-folder in my local git repository. You don't need a zip, Kodi just executes what's in the addon folder.
Testing Kodi in different Virtualbox environments is more fun for me than writing unit test classes.

Issue #52, opened 2019-01-13
Subtitles always shown even when disabled in the add-in


peno64:
**Describe the bug**
For example "Het journaal" always shows subtitles (with an extra empty line as described in another issue) even when in settings "Show subtitles if available" is off.

**To Reproduce**
1. A-Z
2. "Het journaal"
3. play any (I do one on 19h)

**Expected behavior**
Since I set "Show subtitles if available" to off, I do not expect subtitles but they are shown.
However when I go to kodi settings, Player, Language, "Preferred subtitle language" which is on Dutch and I change this to None then no subtitles are shown.

**(please complete the following information):**
 - Operating system: LibreELEC with Kodi 18 latest release candidate
 - Kodi Version: 18

**Additional context**
- Using a VPN: yes (BE)
- Country you are using the addon from: Belgium

**Log (if available)**

mediaminister:
Thanks for reporting. I investigated this problem and I made an experimental fix. Source code: https://github.com/mediaminister/plugin.video.vrt.nu/commit/cbcc2fee2fc9a746d4467fc3db0d7ee13257d5da
You can test this fix with the following zip:

[plugin.video.vrt.nu.zip](https://github.com/pietje666/plugin.video.vrt.nu/files/2753575/plugin.video.vrt.nu.zip)

Please report if this fix works for you.

peno64:
Well the subtitles are now gone, but also when I set "Show subtitles if available" to on...
In kodi settings, Player, Language, "Preferred subtitle language" is on Dutch ...
So it looks like subtitles are now never shown.

mediaminister:
Which program did you try with? VRT doesn't offer subtitles during the first hours after a live broadcast of "Het Journaal": https://www.vrt.be/vrtnu/a-z/het-journaal/2019/het-journaal-het-journaal-19u-20190114/ It seems like VRT is adding subtitles for live broadcasted current affairs programs manually.

Please try with "Het Journaal 13u", there are subtitles available.

peno64:
ah. will try this in a small hour

peno64:
Indeed you are correct.
It works fine now

mediaminister:
Thanks for confirming, I made a pull request with the fix.

mediaminister:
This can be closed as the pull request is merged

PR #54, opened 2019-01-15
bugfix show subtitles addon setting

mediaminister:
Pull request for https://github.com/pietje666/plugin.video.vrt.nu/issues/52

### DIFF
--- a/plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
+++ b/plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
@@ -51,13 +51,16 @@ class KodiWrapper:
             play_item.setProperty('inputstream.adaptive.license_key', video.license_key)
 
         if self.get_setting('showsubtitles') == 'true':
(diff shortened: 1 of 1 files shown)


pietje666:
`         subtitles_visible = False
        if self.get_setting('showsubtitles') == 'true' and video.subtitle_url is not None:
            subtitles_visible = True
            play_item.setSubtitles([video.subtitle_url])

        xbmcplugin.setResolvedUrl(self._handle, True, listitem=play_item)
        while not xbmc.Player().isPlaying():
            xbmc.sleep(10)
        xbmc.Player().showSubtitles(subtitles_visible)`

Can it be simplified to this? Or is the play_item.setSubtitles([video.subtitle_url]) only in certain cases?

mediaminister:
setSubtitles is used for a separate WEBVTT-url together with a HLS-url for compatibility with Kodi 17.3 without inputstream.adaptive. It's also used for HLS_AES when DRM is disabled.

So, for compatibility reasons, it can't be simplified.

Issue #55, opened 2019-01-18
Categories no longer works


dagwieers:
**Describe the bug**
Probably a parsing bug because the website changed again.

**To Reproduce**
1. Go to 'Categories'
2. Click on '.Categories'
3. See empty listing

**Expected behavior**
See the category list.

**(please complete the following information):**
 - LibreELEC v8.95.002
 - Kodi v18.0-RC4

**Log (if available)**
There is no error in the logs.

pietje666:
The site gives back empty categories as well. So its a vrt nu issue. But i will leave this open until it is working again. I wonder if they bring it back up if it will still works the same.

michaelarnauts:
The mobile app seems to be fine. Maybe the app has a more stable API that we can use?

pietje666:
> The mobile app seems to be fine. Maybe the app has a more stable API that we can use?

I would be glad to use one, i decompiled the app once but did not find out how it gets its data. If someone has more info about an Api and how to call it it would be great :)

mediaminister:
You can use the search api:

For example: https://search.vrt.be/search?facets[categories]=koken&buckets[programs]=dagelijkse-kost

dagwieers:
That information is not available from https://services.vrt.be ?

pietje666:
fixed in #56

pietje666:
for api discussions i created #57 so elaborate any further info there :)

PR #56, opened 2019-01-19
fixed categories and repaired legacy support

mediaminister:
Quick fix for https://github.com/pietje666/plugin.video.vrt.nu/issues/55

### DIFF
--- a/plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
+++ b/plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
@@ -51,9 +51,11 @@ class KodiWrapper:
             play_item.setProperty('inputstream.adaptive.license_key', video.license_key)
         
         subtitles_visible = False
-        if self.get_setting('showsubtitles') == 'true' and video.subtitle_url is not None:
+        if self.get_setting('showsubtitles') == 'true':
             subtitles_visible = True
-            play_item.setSubtitles([video.subtitle_url])
+            #separate subtitle url for hls-streams
+            if video.subtitle_url is not None:
+                play_item.setSubtitles([video.subtitle_url])
 
         xbmcplugin.setResolvedUrl(self._handle, True, listitem=play_item)
         while not xbmc.Player().isPlaying():
--- a/plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
+++ b/plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
@@ -74,6 +74,9 @@ class KodiWrapper:
     def open_settings(self):
         self._addon.openSettings()
 
+    def has_inputstream_adaptive_installed(self):
+        return xbmc.getCondVisibility('System.HasAddon("{0}")'.format('inputstream.adaptive')) == 1
+
     def has_widevine_installed(self):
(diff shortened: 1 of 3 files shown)


plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
```diff
@@ -72,6 +74,9 @@ def get_setting(self, setting_id ):
     def open_settings(self):
         self._addon.openSettings()
 
+    def has_inputstream_adaptive_installed(self):
```
pietje666: you cannot install the addon without inputstream adaptive since its a prerequisite in the addon.xml

plugin.video.vrt.nu/resources/lib/vrtplayer/urltostreamservice.py
```diff
@@ -152,8 +152,10 @@ def _select_stream(self, stream_dict, vudrm_token):
             return self._try_get_drm_stream(stream_dict, vudrm_token)
         elif vudrm_token:
             return streamurls.StreamURLS(*self._select_hls_substreams(stream_dict['hls_aes']))
-        else:
+        elif self._kodi_wrapper.has_inputstream_adaptive_installed():
```
pietje666: So because you cannot install version 1.5.+ without having inputstream adaptive installed this is unessecary

plugin.video.vrt.nu/resources/lib/vrtplayer/vrtplayer.py
```diff
@@ -231,6 +236,15 @@ def __get_az_thumbnail_and_title(element):
...
+    def __get_category_thumbnail_and_title(element):
+        thumbnail = VRTPlayer.__format_category_image_url(element)
+        found_element = element.find('h2')
+        title = ''
+        if found_element is not None:
+            title = statichelper.replace_newlines_and_strip(found_element.contents[0])
+        return thumbnail, title
     
     @staticmethod
     def __get_thumbnail_and_title(element):
```
pietje666: I gues this should be removed?

plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
mediaminister: I know, but you can uninstall inputstream.adaptive in Kodi when the VRT NU addon is disabled. I added this in the first place to preserve support for the non-drm hls streams. Just to be prepared for a future worst case scenario when all other streams fail to play.

plugin.video.vrt.nu/resources/lib/vrtplayer/urltostreamservice.py
mediaminister: I know, this is just to preserve support for the non-drm hls streams. You never know if the drm protected streams will fail some point in the future

plugin.video.vrt.nu/resources/lib/vrtplayer/vrtplayer.py
mediaminister: You're right. I have removed it.

plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
pietje666: Will merge but add a comment, else its confusing

PR #58, opened 2019-01-27
bugfix missing license_key

mediaminister:


pietje666:
thanks, still had to investigate why it did not play :)

mediaminister:
Okay, maybe using keyword arguments here is not the best coding practice. If it's less confusing for you, you can also write separate class functions for drm and non drm streamurls.

pietje666:
I think its fine!

PR #68, opened 2019-03-15
Fix the broken items from pylint

dagwieers:
Don't merge this yet until I fixed the remaining things.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,7 @@ language: python
 python:
 - '2.7'
 - '3.6'
-- '3.7'
-- '3.8-dev'
+- '3.7-dev'
 
 sudo: false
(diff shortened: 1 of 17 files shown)


dagwieers:
@pietje666 I fixed everything or ignored a set of known issues, now I still have to test if the addon still works correctly :-)

How do you run the unit tests ? It would be nice to integrate this as well.

dagwieers:
Also, I would like to have a `make test` target, so we can easily perform all these tests on the command line.

dagwieers:
And if you enable Travis, we can see the status of our build, and every PR individually.

dagwieers:
Damn, you merged, just before I added the unit tests :-)

PR #69, opened 2019-03-15
 Add unit tests

dagwieers:
This PR includes:
- Running unit tests from Travis
- Improvements to Makefile to add `make unittest`
- A requirements.txt file (this also makes GitHub make dependencies with other projects)
- A Travis badge on the README
- 2 fixes for Python 3 compatibility (wrt. urlencode and urljoin)
- A fix that broke the addon (related to use_inputstream_adaptive)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,15 +12,21 @@ zip_dir = $(name)/
 
 .PHONY: test
 
-
 package: zip
 
-test:
+test: unittest
 	@echo -e "\e[1;37m=\e[1;34m Starting tests\e[0m"
 	tox -e py27,py36
 	pylint $(name)/*.py
 	pylint $(name)/resources/lib/*/*.py
 	@echo -e "\e[1;37m=\e[1;34m Tests finished successfully.\e[0m"
+
+unittest:
+	@echo -e "\e[1;37m=\e[1;34m Starting unit tests\e[0m"
+	PYTHONPATH=$(name) python $(name)/vrtnutests/apihelpertests.py
+	PYTHONPATH=$(name) python $(name)/vrtnutests/urltostreamservicetests.py
+	PYTHONPATH=$(name) python $(name)/vrtnutests/vrtplayertests.py
+	@echo -e "\e[1;37m=\e[1;34m Unit tests finished successfully.\e[0m"
 
 zip:
 	@echo -e "\e[1;37m=\e[1;34m Building new package\e[0m"
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,11 @@ zip_dir = $(name)/
 
 package: zip
 
+clean:
+	@echo -e "\e[1;37m=\e[1;34m Clean up project directory\e[0m"
+	find . -name '*.pyc' -delete
+	@echo -e "\e[1;37m=\e[1;34m Finished cleaning up.\e[0m"
+
 test: unittest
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -12,16 +12,22 @@ zip_dir = $(name)/
...
 	@echo -e "\e[1;37m=\e[1;34m Starting tests\e[0m"
 	tox -e py27,py36
 	pylint $(name)/*.py
 	pylint $(name)/resources/lib/*/*.py
 	@echo -e "\e[1;37m=\e[1;34m Tests finished successfully.\e[0m"
 
+unittest:
+	@echo -e "\e[1;37m=\e[1;34m Starting unit tests\e[0m"
+	PYTHONPATH=$(name) python $(name)/vrtnutests/apihelpertests.py
+	PYTHONPATH=$(name) python $(name)/vrtnutests/urltostreamservicetests.py
```
pietje666: I think its better to remove the urltostreamservicetests since they will only work property when logged in with an account.

dagwieers: All three work fine on my system just like that ?

dagwieers: Aha, I am not really testing anything. So like I asked you before, how should I invoke these tests ?
These are only classes.

pietje666:
Actually im not sure if we should add the unit tests in the build or not since all the tests are prone to be broken since it are more integration tests

dagwieers:
@pietje666 How do I run the unit tests ?

pietje666:
I always run them through my IDE Visual Studio => but as i said most are integration tests, testing the scraping functionality/loggin mechanism. Hence livestream tests wil never work since it needs an account to be filled in. When developping i use a vrtnu test account but i do not check it in into github ofcourse. Api helper tests are also broken due to a change in episode numbers.

dagwieers:
I do think we need to have these unit tests in Travis. But not the stuff that needs credentials. If we're not testing our code by running it and performing function calls we don't know what works and if stays working when improving.

dagwieers:
Since I still don't know how you actually ran those unit tests (since they cannot be executed) I made them work the only way I know how to. If you end up finding how you run them, we can remove those provisions again.

dagwieers:
So I think this is finally ready to be merged.
However, we may want to fix a few of the tests we ignored, and we need more unit tests.

And for Python3 support: *the proof of the pudding is in the eating.*

dagwieers:
Now also tested in "production" :-)

plugin.video.vrt.nu/resources/lib/helperobjects/streamurls.py
```diff
@@ -11,7 +11,7 @@ def __init__(self, stream_url, subtitle_url=None, license_key=None, use_inputstr
         self._stream_url = stream_url
         self._subtitle_url = subtitle_url
         self._license_key = license_key
-        self._use_inputstream_adaptive = use_inputstream_adaptive
+        self.use_inputstream_adaptive = use_inputstream_adaptive
```
pietje666: is this correct? Shouldn't this be like _use_inputstream_adaptive  this? like the others

pietje666: nvm saw where it is comming from and corrected it

pietje666:
need to get used to travis as well merged too quickly after commit

dagwieers:
You can configure Github that PRs cannot be merged if they don't flag green. (Although you can force it anyway if needed)

PR #71, opened 2019-03-18
A few more PY3 fixes

### LINKED ISSUE #66: Migration to Python 2/3-compatible addon
In this issue we can discuss  the migration proces to a Python 2-3 compatible addon.

I think we definitely need to add the kodi.six library. Any ideas on this? https://github.com/romanvm/kodi.six

Secondly we need to import the Future library like this:
```
#!/usr/bin/env python
# -*- coding: UTF-8 -*-

from future import standard_library
standard_library.install_aliases()
from builtins import str
```

We can look at https://github.com/skipmodea1/plugin.video.tweakers as an example

dagwieers:
This PR also makes the following changes:
- Ensure we import either the Python3 or Python2 libraries for `urljoin()` and `parse_qsl`
- Make the tox environments configurable when using `make test`.
  One can now do: `make test ENVS=py26,py27,py36,py37`
- Fix issues I introduced when removing access to private variables
- Run the tests before building a ZIP file (to avoid future issues like the previous one)

With this PR, [a Python3 *BeautifulSoup* package](https://mirrors.kodi.tv/addons/migration/script.module.beautifulsoup4/) and a newer *inputstreamhelper* package (or some fixes) the VRT.NU addon works on Kodi 18.2 with native Python3 !

This fixes #66

### DIFF
--- a/plugin.video.vrt.nu/resources/lib/vrtplayer/tokenresolver.py
+++ b/plugin.video.vrt.nu/resources/lib/vrtplayer/tokenresolver.py
@@ -26,7 +26,7 @@ class TokenResolver:
 
     def get_ondemand_playertoken(self, token_url, xvrttoken):
         token_path = self._kodi_wrapper.get_userdata_path() + self._ONDEMAND_COOKIE
-        token = self._get_cached_token(token_path)
+        token = self._get_cached_token(token_path, 'vrtPlayerToken')
 
         if token is None:
             cookie_value = 'X-VRT-Token=' + xvrttoken
@@ -36,7 +36,7 @@ class TokenResolver:
 
     def get_live_playertoken(self, token_url):
         token_path = self._kodi_wrapper.get_userdata_path() + self._LIVE_COOKIE
-        token = self._get_cached_token(token_path)
+        token = self._get_cached_token(token_path, 'vrtPlayerToken')
         if token is None:
             headers = {'Content-Type': 'application/json'}
             token = TokenResolver._get_new_playertoken(token_path, token_url, headers)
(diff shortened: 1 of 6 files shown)


plugin.video.vrt.nu/resources/lib/vrtplayer/tokenresolver.py
```diff
@@ -77,26 +83,26 @@ def _get_new_xvrttoken(self, path, get_roaming_token):
...
-            cookie_jar = requests.post(self._TOKEN_GATEWAY_URL, headers=headers, json=payload).cookies
-
-            xvrttoken = TokenResolver._create_token_dictionary(cookie_jar)
-            token = xvrttoken['X-VRT-Token']
-            if get_roaming_token:
-                xvrttoken = TokenResolver._get_roaming_xvrttoken(login_cookie, xvrttoken)
-                token = xvrttoken['X-VRT-Token'] if xvrttoken is not None else None
-            json.dump(xvrttoken, open(path, 'w'))
-        else:
+        if logon_json['errorCode'] != 0:
```
pietje666: With this implementation when a login fails it will show an error message. And afterwards the method continues without being logged in. Which will most likely result in an error. Am i missing something?

dagwieers: You're right, there was a missing `return None` statement.

dagwieers: Fixed.

dagwieers: I removed that change, because TBH it is not really important at all.

PR #82, opened 2019-03-21
Add fanart to menus

dagwieers:
This PR adds directory icons and fanart to all menus.

Beware that the fanart is the same as the thumbnail, which for categories, tvshows and main menu could be done better (by using the parent's thumbnail, or no fanart at all).

But even with the thumbnail as fanart, IMO it looks better.

Also the icons do not work at the moment...


dagwieers:
So while we should improve the fanart used, and make the icons work, this PR is better than no PR.

PR #84, opened 2019-03-21
Restore the channel thumbnail to text-based thumbnail

dagwieers:
Since we now have the live stream screenshot as fanart, we can safely turn back to text-based icons for channels, so that adding it to favourites works properly again.


dagwieers:
I hope we can restore the official channel icons properly some day though.

PR #86, opened 2019-03-21
More visual improvements

dagwieers:
This PR includes:
- More icons and fanart added
- Simplified artwork using art_dict
- Add more metadata to tvshows and episodes
- Indicate in plot if geo-blocked
- Indicate when a video will disappear in the next 3 months
- Rename replace_newlines_and_strip() to strip_newlines()
- Rename replace_double_slashes_with_https() to add_https_method()
- Improve add_https_method to simply do the right thing
- Turn args calls into kwargs calls for TitleItem

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,9 @@ you can send a message to [our Facebook page](https://www.facebook.com/kodivrtnu
 
 ## Releases
 
-#### v1.7.0 (2019-03-21)
+#### v1.7.0 (2019-03-23)
+- Indicate when content will disappear in the next 3 months (@dagwieers)
+- Indicate when content is geo-blocked (@dagwieers)
 - Add fanart to menus (@dagwieers)
 - Fix issue related to missing sound (@mediaminister)
 - Improve main menu listing (@dagwieers)
--- a/plugin.video.vrt.nu/addon.xml
+++ b/plugin.video.vrt.nu/addon.xml
@@ -25,7 +25,9 @@
         <platform>all</platform>
         <license>GNU General Public License, v3</license>
         <news>
-v1.7.0 (2019-03-21)
+v1.7.0 (2019-03-23)
+- Indicate when content will disappear in the next 3 months (@dagwieers)
+- Indicate when content is geo-blocked (@dagwieers)
 - Add fanart to menus (@dagwieers)
 - Fix issue related to missing sound (@mediaminister)
 - Improve main menu listing (@dagwieers)
--- a/plugin.video.vrt.nu/resources/language/resource.language.en_gb/strings.po
+++ b/plugin.video.vrt.nu/resources/language/resource.language.en_gb/strings.po
@@ -119,3 +119,14 @@ msgctxt "#32104"
 msgid "Sporza"
 msgstr "Sporza"
 
+msgctxt "#32201"
+msgid "[B][COLOR red]Geo-blocked[/COLOR][/B]\n"
+msgstr "[B][COLOR red]Geo-blocked[/COLOR][/B]\n"
+
+msgctxt "#32202"
+msgid "[B][COLOR blue]Available until %s[/COLOR][/B]\n"
+msgstr "[B][COLOR blue]Available until %s[/COLOR][/B]\n"
+
+msgctxt "#32203"
+msgid "[B][COLOR green](%s days left)[/COLOR][/B]\n"
+msgstr "[B][COLOR green](%s days left)[/COLOR][/B]\n"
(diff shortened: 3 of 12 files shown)


dagwieers:
@pietje666 @mediaminister Can you please test this and give feedback.

Even though it looks like a lot of changes, it is mostly restructuring a few things.

dagwieers:
We will now show the following information in the plot:
> **Geo-Blocked**
> Available until 2019-05-28
> (68 days left)

To make it very convenient for users.

dagwieers:
@pietje666 @mediaminister Now that we removed most of the scraping, we also have to be careful when trusting some data is available. For instance we do things like: `tvshow['broadcastDate']`, but this will fail if this information is no longer (or accidentally not) available. If we do things like `tvshow.get('broadcastDate')` it will not fail but be set to None.

In itself that is better, but in some cases we need to handle None values better. In any case, for VRT supplied data we cannot assume the format is consistent, or set into stone. This PR uses `get()`.

dagwieers:
If you could evaluate this PR we can maybe get this into v1.7.0 as well.

mediaminister:
I added two comments to the code:
- localized strings and date formats
- shortDescription data should be "beautifulsouped"

mediaminister:
Crash:
```
2019-03-21 20:53:26.647 T:139830741829376   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                            Error Type: <type 'exceptions.TypeError'>
                                            Error Contents: attribute of type 'NoneType' is not callable
                                              File "/home/mediaminister/.kodi/addons/plugin.video.vrt.nu/addon.py", line 57, in <module>
                                                router(sys.argv[2][1:])
                                              File "/home/mediaminister/.kodi/addons/plugin.video.vrt.nu/addon.py", line 43, in router
                                                vrt_player.show_episodes(params['video_url'])
                                              File "/home/mediaminister/.kodi/addons/plugin.video.vrt.nu/resources/lib/vrtplayer/vrtplayer.py", line 89, in show_episodes
                                                episode_items, sort = self._api_helper.get_episode_items(path)
                                              File "/home/mediaminister/.kodi/addons/plugin.video.vrt.nu/resources/lib/vrtplayer/vrtapihelper.py", line 73, in get_episode_items
                                                episode_items, sort_method = self._map_to_episode_items(api_json['results'])
                                              File "/home/mediaminister/.kodi/addons/plugin.video.vrt.nu/resources/lib/vrtplayer/vrtapihelper.py", line 107, in _map_to_episode_items
                                                metadata_creator.ontime = datetime.strptime(episode.get('assetOnTime'), '%Y-%m-%dT%H:%M:%S+0000')
                                            TypeError: attribute of type 'NoneType' is not callable
                                            -->End of Python script error report<--
```

dagwieers:
For the localized string, I first want to determine what we want here. If everybody is happy with how it is now, I will add the translations. That bug I hadn't seen, but I was expecting things like this (see previous comment).

dagwieers:
Hmm, in this case, I did test for None values:
```
            if episode.get('assetOnTime') is not None:
                metadata_creator.ontime = datetime.strptime(episode.get('assetOnTime'), '%Y-%m-%dT%H:%M:%S+0000')
```
But I will test later tonight, playing indoor soccer now. Please let me know for which content you are seeing this, will help in testing later :-)

mediaminister:
I'll have a fix soon

mediaminister:
Oh no, we hit nasty a Kodi/Python bug: https://forum.kodi.tv/showthread.php?tid=112916
Fix:
```python
            if episode.get('assetOnTime') is not None:
                metadata_creator.ontime = datetime(*(time.strptime(episode.get('assetOnTime'), '%Y-%m-%dT%H:%M:%S+0000')[0:6]))
            if episode.get('assetOffTime') is not None:
                metadata_creator.offtime = datetime(*(time.strptime(episode.get('assetOffTime'), '%Y-%m-%dT%H:%M:%S+0000')[0:6]))
```

dagwieers:
So this is a Windows-specific issue ? At first I tried implementing this using time only, but it wasn't as obvious to get proper time deltas. You basically have to convert to epoch times, subtract and compare.

Shall I make a helper-function for this ?

dagwieers:
@mediaminister IMO we ought to learn the world the ISO standard for dates %Y-%m-%d :-)

But we really ought to be using the locale date formats from Kodi. But how do we get to those ?

dagwieers:
Alright, strings and date-formats are now localized!

mediaminister:
About ISO date: Kodi provides localisation and the user can choose this option. So, in fact the user decides which date format is shown.

It doesn't crash anymore, so it's almost ready for release.
Still a couple of little bugs:
- html tags in shortdescription should be escaped
- 0 days left should be ... hours left

dagwieers:
@mediaminister I am using the localized date format. Works fine. I also
discovered how to access global settings (using JSON-RPC) so expect
improved proxy-support soon.

PR #87, opened 2019-03-22
Add full proxy support

dagwieers:
This PR gets the Kodi global proxy settings and then applies those to
all invocations of requests.

This PR also prefixes log messages using `[plugin.video.vrt.nu] ` so it is clear where log messages are coming from.

This is the first implementation that adds full proxy support.

In the future I would like to add 2 additional features so that users
can select between:

- Use Kodi proxy settings (default)
- Disable proxy support for this addon
- Custom proxy settings (for this addon)

The current PR implements the first (default) option.

### DIFF
--- a/plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
+++ b/plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
@@ -24,6 +24,7 @@ class KodiWrapper:
         self._handle = handle
         self._url = url
         self._addon = addon
+        self._addon_id = addon.getAddonInfo('id')
 
     def show_listing(self, list_items, sort=None, content_type='episodes'):
         listing = []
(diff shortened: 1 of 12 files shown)


dagwieers:
@mediaminister @pietje666 Please review!

PS I updated [the wiki page with ideas](https://github.com/pietje666/plugin.video.vrt.nu/wiki/Ideas), feedback welcome.

dagwieers:
While testing (easiest is to remove your default gateway when the proxy is in your local network) I discovered that **streaminputhelper** is not proxy-aware either. So DRM-enable content (i.e. live streams) fail to work because `No route to host` when contacting `dl.google.com`.

But our addon at least is using the proxy fully.

plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
```diff
@@ -91,6 +92,53 @@ def get_setting(self, setting_id):
...
+        usehttpproxy = self.get_global_setting('network.usehttpproxy')
+        if usehttpproxy is False:
+            return dict()
+
+        httpproxytype = self.get_global_setting('network.httpproxytype')
+
+        if httpproxytype != 0:
+            title = self.get_localized_string(32061)
+            message = self.get_localized_string(32062)
+            self.show_ok_dialog(title, message)
```
mediaminister: Please return None if it doesn't work. Don't let Kodi crash.

dagwieers: Ok.

dagwieers: The problem is that I don't know if **requests[socks]** is installed. So at least now it would work if it is installed. I opened a ticket to get this support added to the official **script.module.requests** package, I think that's the best option we have.

PS I am also adding this to other addons now, planning to add it to netflix as well. So hopefully the need for SOCKS support will increase due to this.

dagwieers: I am going to merge as-is, we can add the SOCKS detection later.

PR #89, opened 2019-03-23
hls streams: support virtual subclips (vbegin and vend parameters)

mediaminister:
VRT makes use of virtual subclips to provide live broadcast current affairs programs in an "on demand" context. For instance: "Het Journaal" in the "Most recent" menu uses this feature.
This fixes availability of current affair programs during and immediately after the live broadcast for users who don't use inputstream.adaptive

### DIFF
--- a/plugin.video.vrt.nu/resources/lib/vrtplayer/streamservice.py
+++ b/plugin.video.vrt.nu/resources/lib/vrtplayer/streamservice.py
@@ -200,7 +200,7 @@ class StreamService:
             direct_audio_url = match_audio[len(match_audio)-1]
 
         # Get video uri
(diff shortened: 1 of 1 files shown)


dagwieers:
I don't know a lot about this, leave it up to @pietje666 to decide :-)

pietje666:
I do not have any knowledge about this neither. As a matter in fact my knowledge about streams is limited!

dagwieers:
Then let's trust the guy who seemingly does :-)

dagwieers:
@pietje666 Only 25 minutes in order to release on 2019-03-23 (as stated in the Changelog ;-))

pietje666:
Im not home :D

dagwieers:
Ok, the next person adding a feature can update the Changelog ;-)

PR #90, opened 2019-03-23
Refactor package maker

pietje666:
Refactored Packagemaker and provided easy way to create zip files for end users.


packagemaker/Program.cs
```diff
@@ -10,25 +10,37 @@ namespace plugin.video.vrt.nu.packagemaker
 {
     class Program
     {
+        private const string ZipfileName = "plugin.video.vrt.nu.zip";
```
dagwieers: IMO it would be nice if this included a version number (and maybe even a git hash) so a new build is unique and somewhat identifiable.

dagwieers:
Who am I not to agree with this ;-)

pietje666:
Will move it to packagemaker later, was a good suggestion! And will try to add version number later.

PR #92, opened 2019-03-23
display hours if episode is less than 1 day available

mediaminister:


plugin.video.vrt.nu/resources/lib/vrtplayer/vrtapihelper.py
```diff
@@ -115,9 +115,12 @@ def _map_to_episode_items(self, episodes, titletype=None):
             if metadata_creator.geolocked:
                 plot_meta += self._kodi_wrapper.get_localized_string(32201)
             # Only display when a video disappears if it is within the next 3 months
-            if metadata_creator.offtime is not None and (metadata_creator.offtime - datetime.now()).days < 92:
+            if metadata_creator.offtime is not None and (metadata_creator.offtime - datetime.utcnow()).days < 92:
```
dagwieers: Good catch. I wasn't that worried about timezones since it would only be 1 hour or 2 hours off and we only used days before. But here it is more important to get it right.

PR #95, opened 2019-03-26
fixed categories

mediaminister:
VRT changed something on their website


dagwieers:
Confirmed current was broken, and this fixes it.

@pietje666 We need to release ASAP or backport to release v1.6.1.

PR #99, opened 2019-03-26
kodi addon checker

mediaminister:
Changes to meet kodi-addon-checker test at https://github.com/xbmc/repo-plugins/pulls

### DIFF
--- a/plugin.video.vrt.nu/addon.xml
+++ b/plugin.video.vrt.nu/addon.xml
@@ -5,7 +5,7 @@
        provider-name="Martijn Moreel">
 
     <requires>
-        <import addon="inputstream.adaptive" version="2.0.19" />
+        <import addon="inputstream.adaptive" version="2.0.19"/>
         <import addon="script.module.beautifulsoup4" version="4.5.3"/>
         <import addon="script.module.inputstreamhelper" version="0.3.3"/>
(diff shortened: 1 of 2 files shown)


plugin.video.vrt.nu/addon.xml
```diff
@@ -16,9 +16,7 @@
     <extension point="xbmc.python.pluginsource" library="addon.py">
         <provides>video</provides>
     </extension>
-
-    <extension point="xbmc.service" library="service.py" start="startup" />
```
dagwieers: Do we remove `start="startup"` here on purpose ?

mediaminister: Yes, cause this option doesn't exist anymore in Kodi Leia: look at the bottom of the page at https://kodi.wiki/view/Service_add-ons

dagwieers: Ok, but it does for Krypton ? Because we still release for Krypton and Leia.

dagwieers: Ok, so I guess we don't need it even when it meant something ?

mediaminister: Correct, we don't need it. Everything works as expected without it.

PR #100, opened 2019-03-26
Include LICENSE and README.md in release

dagwieers:
So next release will be easier ;-)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ git_hash = $(shell git rev-parse --short HEAD)
 zip_name = $(name)-$(version)-$(git_hash).zip
 exclude_files = $(name).pyproj vrtnutests/ vrtnutests/*
 exclude_paths = $(patsubst %,$(name)/%,$(exclude_files))
+include_files = LICENSE README.md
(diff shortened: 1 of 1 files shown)


dagwieers:
No windows ;-)

dagwieers:
If we merge #102 this is actually no longer needed.

pietje666:
> No windows ;-)

No windows needed!  Although it would have crashed at the Process.Start (that's why i removed it in the PR). Actually first i was making a self contained binary but it was 70 mb for a packager.

https://dotnet.microsoft.com/download/linux-package-manager/ubuntu18-04/runtime-2.1.2

dagwieers:
Don't take this the wrong way, but creating a zip file is just a simple command.
Running make is so much easier, and available on almost every Linux system.

I know the Microsoft world is adopting Linux at a fast pace, and I have been doing Powershell stuff the past few years (for Ansible), but I like to keep simple things simple, and for me that excludes using .Net core for creating ZIP files. Sorry :-)

pietje666:
I understand but i am bad at scripting!

PR #101, opened 2019-03-26
Router simplify

dagwieers:
I had this one ready before, but didn't want to slow down any release.


mediaminister:
Looks good to me, but why is this safer?

dagwieers:
@mediaminister Works also if `action` would be missing from the dictionary (I guess that may be hypothetical ?).

PR #111, opened 2019-03-28
Improve GitHub templates, add feature/question

dagwieers:
On the Wiki we ask for ideas, so why not provide entry templates for this as well?


dagwieers:
Merging in order to test. Feel free to discuss or propose improvements.

PR #115, opened 2019-03-29
Improve thumbnail/fanart information

dagwieers:
Currently we use the same thumbnail and fanart for video listings, but we can actually show something different in most cases. So that the background shows the programImageUrl and it sticks around when browsing the episodes.

This also adds fanart to season listings.

This PR also increases the number of recent items to 100.


dagwieers:
This closes 2 items on our wishlist !

mediaminister:
Looks good to me.

PR #117, opened 2019-03-30
Make code more robust

### LINKED ISSUE #35: Idea: Implement more resilience into the scraper/parser
**Describe the bug**
It's unfortunate that there is no better API to VRT.NU (especially since we all pay for this service anyway and the VRT is ending DVB-T free-to-air soon).

But that doesn't mean we cannot improve the experience for end-users:

- If the parser fails, dump the payload and URL so people have an easier time understanding what we expect and what changed, and offer a fix.

- If the parser fails, on non-critical metadata (and I guess most is) we could replace it with a stub identifier, but still offer the video's. In this case the end-user may not see e.g. the titles, but can still select a video to play

- If the parser fails, provide a warning (in case we do continue) or an error (in case we cannot) with some more information:

  - *This issue is caused by the VRT changing the web interface (the only interface to VRT.NU that currently exists).*

  - *Please check http://github.com/pietje666/plugin.video.vrt.nu/issues for known issues and fixes.*

  - *Or investigate the logs to fix this yourself and contribute the fix.*

- Maybe we can have a better/faster distribution mechanism for this add-on. E.g. we can create our own repository that would automatically get the latest version. Which means if we release a new version, it could be a matter of minutes/hours to have Kodi update the add-on, rather than days.

dagwieers:
Make every dictionary lookup use get() which doesn't raise an exception
when missing. (If JSON returns something unexpected, or web scraping
fails unexpectedly)

We still have to look for each dictionary lookup what we should return
if the lookup fails. By default it returns None, which sometimes is
fine, but in some cases we better return an empty list or an empty dict.

Some parts now will undoubtedly fail because of None values where it
would have failed Kodi before.

In some cases a missing title would be better replaced with e.g. 'N/A'
or 'video #X' so the listing would still work. Either a complete review
is necessary, or future brokeness will guide us to make better decisions
;-)

This fixes #35

### DIFF
--- a/plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
+++ b/plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
@@ -107,7 +107,7 @@ class KodiWrapper:
 
     def get_global_setting(self, setting):
         json_result = xbmc.executeJSONRPC('{"jsonrpc": "2.0", "method": "Settings.GetSettingValue", "params": {"setting": "%s"}, "id": 1}' % setting)
-        return json.loads(json_result)['result']['value']
+        return json.loads(json_result).get('result', dict()).get('value')
 
     def get_proxies(self):
         usehttpproxy = self.get_global_setting('network.usehttpproxy')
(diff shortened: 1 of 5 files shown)


pietje666:
using the Api i doubt there will be alot of missing titles! But still seems a good practice

PR #118, opened 2019-03-31
WIP: Add more metadata

dagwieers:
This PR includes:
- Add 'score' as 'rating'
- Add 'genre'
- ~Remove unused MetadataCollector: class~

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -16,6 +16,7 @@ disable=
     too-many-branches,
     too-many-instance-attributes,
     too-many-locals,
+    too-many-public-methods,
     too-many-statements,
     unnecessary-lambda,
     unused-argument,
--- a/resources/lib/kodiwrappers/kodiwrapper.py
+++ b/resources/lib/kodiwrappers/kodiwrapper.py
@@ -74,9 +74,10 @@ class KodiWrapper:
 #            xbmcplugin.setProperty(handle=self._handle, key='sort.order', value=str(sort_methods['unsorted']))
 
         for title_item in list_items:
-            list_item = xbmcgui.ListItem(label=title_item.title, thumbnailImage=title_item.art_dict.get('thumb'))
+            list_item = xbmcgui.ListItem(label=title_item.title)
             url = self._url + '?' + urlencode(title_item.url_dict)
             list_item.setProperty(key='IsPlayable', value='true' if title_item.is_playable else 'false')
+            list_item.setProperty(key='video', value='true' if title_item.is_playable else 'false')
 
             # FIXME: This does not appear to be working, we have to order it ourselves
 #            list_item.setProperty(key='sort.ascending', value='true' if ascending else 'false')
@@ -194,7 +195,7 @@ class KodiWrapper:
 
         return dict(http=proxy_address, https=proxy_address)
 
-    # Note: InputStream Adaptive is not pre-installed on Windows and in some cases users can uninstall it
+    # NOTE: InputStream Adaptive is not pre-installed on Windows and in some cases users can uninstall it
     def has_inputstream_adaptive_installed(self):
         return xbmc.getCondVisibility('System.HasAddon("{0}")'.format('inputstream.adaptive')) == 1
 
--- a/resources/lib/vrtplayer/__init__.py
+++ b/resources/lib/vrtplayer/__init__.py
@@ -32,7 +32,7 @@ CHANNELS = {
         name='Eén',
         studio='Een',
         live_stream='https://www.vrt.be/vrtnu/kanalen/een/',
-        live_stream_id='vualto_een',
+        live_stream_id='vualto_een_geo',
     ),
     'canvas': dict(
         id='1H',
@@ -40,7 +40,7 @@ CHANNELS = {
         name='Canvas',
         studio='Canvas',
         live_stream='https://www.vrt.be/vrtnu/kanalen/canvas/',
-        live_stream_id='vualto_canvas',
+        live_stream_id='vualto_canvas_geo',
     ),
     'ketnet': dict(
         id='O9',
@@ -48,7 +48,7 @@ CHANNELS = {
         name='Ketnet',
         studio='Ketnet',
         live_stream='https://www.vrt.be/vrtnu/kanalen/ketnet/',
-        live_stream_id='vualto_ketnet',
+        live_stream_id='vualto_ketnet_geo',
     ),
     'ketnet-jr': dict(
         id='1H',
@@ -61,7 +61,7 @@ CHANNELS = {
         type='radio+tv',
         name='Sporza',
         studio='Sporza',
-        live_stream_id='vualto_sporza',
+        live_stream_id='vualto_sporza_geo',
     ),
     'vrtnxt': dict(
         id='',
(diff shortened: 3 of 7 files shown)


plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
```diff
@@ -51,13 +51,14 @@ def show_listing(self, list_items, sort=None, content_type='episodes'):
         for title_item in list_items:
             list_item = xbmcgui.ListItem(label=title_item.title, thumbnailImage=title_item.art_dict.get('thumb'))
             url = self._url + '?' + urlencode(title_item.url_dict)
-            list_item.setProperty('IsPlayable', str(title_item.is_playable))
+            list_item.setProperty('isPlayable', 'true' if title_item.is_playable else 'false')
```
pietje666: setProperty is not case sensitive but in the kodi examples for setProperty they do start the values with a capital letter. 
from the docs => 

```
listitem.setProperty('AspectRatio', '1.85 : 1')
listitem.setProperty('StartOffset', '256.4')
```

plugin.video.vrt.nu/resources/lib/vrtplayer/vrtapihelper.py
```diff
@@ -31,11 +31,13 @@ def get_tvshow_items(self, path):
...
-            metadata_creator.tvshowtitle = tvshow.get('title')
-            metadata_creator.plot = statichelper.unescape(tvshow.get('description'))
-            metadata_creator.brands = tvshow.get('brands')
+            metadata = metadatacreator.MetadataCreator()
+            metadata.mediatype = 'tvshow'
+            metadata.tvshowtitle = tvshow.get('title')
+            metadata.plot = statichelper.unescape(tvshow.get('description'))
+            metadata.plotoutline = metadata.plot
+            metadata.brands = tvshow.get('brands')
+            metadata.score = tvshow.get('score')  # NOTE: The score here is always between 4.8 and 5.0
```
mediaminister: I think score means the relevance of the search query here, for a different query, you get a different score:
https://vrtnu-api.vrt.be/search?i=video&facets[url]=//www.vrt.be/vrtnu/a-z/inside-europe--ten-years-in-turmoil/2018/inside-europe--ten-years-in-turmoil-s2018a2/

It's not related with a user rating or something else.

plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
```diff
@@ -51,13 +51,14 @@ def show_listing(self, list_items, sort=None, content_type='episodes'):
...
 
             if title_item.art_dict:
                 list_item.setArt(title_item.art_dict)
 
             if title_item.video_dict:
                 list_item.setInfo('video', infoLabels=title_item.video_dict)
+                list_item.setProperty('id', title_item.video_dict.get('videoid'))
```
mediaminister: Is this useful?

mediaminister:
I also get the following errors in my Kodi log:
```
2019-03-31 17:14:35.548 T:140397439346432   ERROR: NEWADDON Unknown Video Info Key "channelname"
2019-03-31 17:14:35.548 T:140397439346432   ERROR: NEWADDON Unknown Video Info Key "enddate"
2019-03-31 17:14:35.548 T:140397439346432   ERROR: NEWADDON Unknown Video Info Key "startdate"
```

plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
dagwieers: Yeah, I wasn't sure and this was just to test something. I looked at other plugins and most were doing it like this. I don't mind turning it back.

Beware this PR is a Work-In-Progress, so should not be merged :-) Some of the things do work correctly, but are not ready, e.g. the score/rating is nice, but can surpass 10, so I need to understand first what the value actually means... For programs the values are between 4.8 and 5.2, for episodes they range from 3 to 8, but I have seen values up to 38... (Undercover)

plugin.video.vrt.nu/resources/lib/vrtplayer/vrtapihelper.py
dagwieers: For programs they are quite low (between 4.8 and 5.1) but for episodes they seem to indicate something else, which I think is popularity or personal preference based on data mining. All the programs I like to watch score very high (sometimes higher than 10), but cycling tours score even higher (which I do not watch, except for Tour de France in the background).

So I don't think they are related just to search relevance, but rather popularity.

plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
dagwieers: This PR is a Work-In-Progress, so I am just testing stuff.

dagwieers:
@mediaminister Ah, nice. These were in the InfoLabel documentation, but I wondered whether the mediatype really affected the InfoLabels or not. I think this proves it :-)

plugin.video.vrt.nu/resources/lib/vrtplayer/vrtapihelper.py
mediaminister: Strange, Gent-Wevelgem has a score of 13.495115 and 44.903103:

https://vrtnu-api.vrt.be/search?i=video&facets[url]=//www.vrt.be/vrtnu/a-z/sporza--wielrennen/2019/sporza--wielrennen-d20190331-gent-wevelgem/
https://vrtnu-api.vrt.be/search?i=video&q=pbs-pub-8ff1a6c9-bdd1-4404-960f-a94a80b0c7b3

dagwieers: Indeed, TBH I think this means that a lot of cycling fans use VRT.NU (e.g. when they are actually live at the event). So the popularity numbers are inflated for cycling because of the sheer number of users who watch it through VRT.NU, whereas most other programs are watch most through broadcasting.

So maybe these outliers are exceptions to the data mining algorithms (in case these are somehow personal recommendations based on profile/viewing habit) or simply a general popularity rating with very few or none personal information (is my profile even used for the listings, I don't think so).

mediaminister: I mean: one single video item has two different scores dependent on the type of api query.

https://www.elastic.co/guide/en/elasticsearch/guide/current/scoring-theory.html

dagwieers: But we didn't provide any keywords, so what would it be scoring against?

dagwieers: BTW I agree that in case this offers no real value, we shouldn't be adding it as score. But I wanted to provoke this kind of discussion by publishing this Work-In-Progress, so thanks in advance.

plugin.video.vrt.nu/resources/lib/kodiwrappers/kodiwrapper.py
dagwieers: BTW I hoped this would somehow influence the "Watched" flag. I should probably start looking into the source code instead...

PR #119, opened 2019-03-31
fix virtual subclips in Kodi

mediaminister:
VRT NU uses [virtual subclips](http://docs.unified-streaming.com/documentation/vod/player-urls.html#virtual-subclips) to provide on demand programs (mostly current affair programs) already from a livestream while or shortly after live broadcasting them. But this feature doesn't work always as expected because Kodi doesn't play the program from the beginning when the ending timestamp of the program is missing from the stream_url. When begintime is present in the stream_url and endtime is missing, we must add endtime to the stream_url so Kodi treats the program as an on demand program and starts the stream from the beginning like a real on demand program.

### DIFF
--- a/resources/lib/vrtplayer/streamservice.py
+++ b/resources/lib/vrtplayer/streamservice.py
@@ -4,6 +4,8 @@
 
 import re
 import requests
+from datetime import datetime, timedelta
+import time
 from bs4 import BeautifulSoup, SoupStrainer
 
 from resources.lib.helperobjects import apidata, streamurls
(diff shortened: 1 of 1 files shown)


pietje666:
Looks good for me, which programs did you have issues with btw?

mediaminister:
This morning all programs on channel 'Eén' were live broadcasted: 'R.K. eredienst', 'De zevende dag', 'Het journaal 13u' and the issue (virtual subclip with missing end time) was present between 10 a.m and 15 a.m.
You can check this issue easily yourself with the "Most Recent" menu every time when a program is live broadcasted.
For example, this afternoon 'Sporza Gent-Wevelgem' "on demand" is not started from the beginning.
And Tonight, 'Het journaal - Sportweekend' will have the same issue.
With this fix, the program is always started from the beginning.

And if you want to watch 'live', you choose the Live-tv menu.

dagwieers:
@mediaminister Nice, I noticed this myself, but didn't know we could actually fix this !

dagwieers:
@mediaminister Sorry, you need to do a rebase to resolve the conflict: `git pull --rebase origin master`

Issue #124, opened 2019-04-05
Same episodes show up in different seasons


dagwieers:
### Describe the bug ###
I noticed this in the seasons of "Universiteit van Vlaanderen" which has 2 seasons listed: **Season 2018-2019** and **Season 2017-2018**. Both seasons have the same episodes listed.

For another Series "Thuis" this is not the case, although it does show episodes released every day for Season 21 and Season 24. (But I guess this is done by the VRT on purpose?)

mediaminister:
For the series "Thuis, VRT broadcasts two episodes a day: one episode form an older season in the afternoon and one new episode in primetime. So, this is normal.

For "Universiteit van Vlaanderen", I fixed this problem by using the `doc_count` value from the seasons buckets as the size parameter and using `seasonName` instead of `seasonTitle` facet.

PR #125, opened 2019-04-05
fixed seasons

mediaminister:
This fixes https://github.com/pietje666/plugin.video.vrt.nu/issues/124

### DIFF
--- a/resources/lib/vrtplayer/vrtapihelper.py
+++ b/resources/lib/vrtplayer/vrtapihelper.py
@@ -87,92 +87,99 @@ class VRTApiHelper:
             else:
                 api_url = path
             api_json = requests.get(api_url, proxies=self._proxies).json()
+            season_key = None
             # Look for seasons items if not yet done
             if 'facets[seasonTitle]' not in path:
                 episode_items = self._get_season_items(api_url, api_json)
+            else:
+                season_key = path.split('facets[seasonTitle]=')[1]
             # No season items, generate episode items
             if not episode_items:
-                episode_items, sort = self._map_to_episode_items(api_json.get('results', []))
+                episode_items, sort = self._map_to_episode_items(api_json.get('results', []), season_key=season_key)
 
         return episode_items, sort
 
-    def _map_to_episode_items(self, episodes, titletype=None):
+    def _map_to_episode_items(self, episodes, titletype=None, season_key=None):
         episode_items = []
         sort = None
         for episode in episodes:
-            display_options = episode.get('displayOptions', dict())
+            # VRT API workaround: seasonTitle facet behaves as a partial match regex,
+            # so we have to filter out the episodes from seasons that don't exactly match.
+            if not (season_key and episode.get('seasonTitle') != season_key):
+                display_options = episode.get('displayOptions', dict())
 
-            if episode.get('programType') == 'reeksoplopend' and titletype is None:
-                titletype = 'reeksoplopend'
+                if episode.get('programType') == 'reeksoplopend' and titletype is None:
+                    titletype = 'reeksoplopend'
 
-            metadata_creator = metadatacreator.MetadataCreator()
-            metadata_creator.tvshowtitle = episode.get('program')
-            if episode.get('broadcastDate') != -1:
-                metadata_creator.datetime = datetime.fromtimestamp(episode.get('broadcastDate', 0) / 1000)
+                metadata_creator = metadatacreator.MetadataCreator()
+                metadata_creator.tvshowtitle = episode.get('program')
+                if episode.get('broadcastDate') != -1:
+                    metadata_creator.datetime = datetime.fromtimestamp(episode.get('broadcastDate', 0) / 1000)
 
-            metadata_creator.duration = (episode.get('duration', 0) * 60)  # Minutes to seconds
-            metadata_creator.plot = BeautifulSoup(episode.get('description'), 'html.parser').text
-            metadata_creator.brands = episode.get('programBrands', episode.get('brands'))
-            metadata_creator.geolocked = episode.get('allowedRegion') == 'BE'
-            if display_options.get('showShortDescription'):
-                short_description = BeautifulSoup(episode.get('shortDescription'), 'html.parser').text
-                metadata_creator.plotoutline = short_description
-                metadata_creator.subtitle = short_description
-            else:
-                metadata_creator.plotoutline = episode.get('subtitle')
-                metadata_creator.subtitle = episode.get('subtitle')
-            metadata_creator.season = episode.get('seasonName')
-            metadata_creator.episode = episode.get('episodeNumber')
-            metadata_creator.mediatype = episode.get('type', 'episode')
-            if episode.get('assetOnTime'):
-                metadata_creator.ontime = datetime(*time.strptime(episode.get('assetOnTime'), '%Y-%m-%dT%H:%M:%S+0000')[0:6])
-            if episode.get('assetOffTime'):
-                metadata_creator.offtime = datetime(*time.strptime(episode.get('assetOffTime'), '%Y-%m-%dT%H:%M:%S+0000')[0:6])
-
-            # Add additional metadata to plot
-            plot_meta = ''
-            if metadata_creator.geolocked:
-                # Show Geo-locked
-                plot_meta += self._kodi_wrapper.get_localized_string(32201)
-            # Only display when a video disappears if it is within the next 3 months
-            if metadata_creator.offtime is not None and (metadata_creator.offtime - datetime.utcnow()).days < 93:
-                # Show date when episode is removed
-                plot_meta += self._kodi_wrapper.get_localized_string(32202) % metadata_creator.offtime.strftime(self._kodi_wrapper.get_localized_dateshort())
-                # Show the remaining days/hours the episode is still available
-                if (metadata_creator.offtime - datetime.utcnow()).days > 0:
-                    plot_meta += self._kodi_wrapper.get_localized_string(32203) % (metadata_creator.offtime - datetime.utcnow()).days
+                metadata_creator.duration = (episode.get('duration', 0) * 60)  # Minutes to seconds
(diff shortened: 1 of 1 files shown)


resources/lib/vrtplayer/vrtapihelper.py
```diff
@@ -166,13 +166,17 @@ def _map_to_season_items(self, api_url, buckets, episode):
...
         for bucket in buckets:
-            season_title = bucket.get('key')
-            title = ''.join((self._kodi_wrapper.get_localized_string(32094), ' ', season_title))
+            season_key = bucket.get('key')
+            season_size = str(bucket.get('doc_count'))
+            title = ''.join((self._kodi_wrapper.get_localized_string(32094), ' ', season_key))
             season_facet = '&facets[seasonTitle]='
-            if ' ' in season_title:
-                season_title = season_title.replace(' ', '-')
+            # Workaround: seasonTitle facet doesn't support spaces and hyphens properly, luckily seasonName facet escapes this
```
dagwieers: To me this is not clear. So I doubt anyone else reading this code will understand why this is needed.
Maybe it would help if we add examples in the comments for each case.

Also, why don't we *always* replace ' ' with '-', and always check for seasonKey instead ?

dagwieers: So my implementation would simply do this:
```python
            season_title = bucket.get('key')
            season_size = str(bucket.get('doc_count'))
            title = ''.join((self._kodi_wrapper.get_localized_string(32094), ' ', season_title))

            # Workaround: seasonTitle facet doesn't support spaces and hyphens properly, luckily seasonName facet escapes this
            season_key = season_title.replace(' ', '-')
            season_facet = '&facets[seasonName]='
```

mediaminister: We get the available season values from faceted search api with `bucket.get('key')`, and **these values always match the `seasonTitle` values**.
For instance: https://vrtnu-api.vrt.be/search?i=video&facets[programUrl]=//www.vrt.be/vrtnu/a-z/girls-talk/
We get value `1`, and `2`
Then we filter the episodes from season 1 with the following request: https://vrtnu-api.vrt.be/search?i=video&facets[programUrl]=//www.vrt.be/vrtnu/a-z/girls-talk/&facets[seasonTitle]=1

We get 5 episodes, this is correct.

And now, try to filter with `seasonName`: https://vrtnu-api.vrt.be/search?i=video&facets[programUrl]=//www.vrt.be/vrtnu/a-z/girls-talk/&facets[seasonName]=1

We get 0 episodes, because the value of `seasonName` is `2018` and doesn't match the `bucket.get('key')` value, wich is `1`.

So, `seasonName` isn't a reliable key to get the episodes from the corresponding season.

dagwieers: Thanks for the detailed explanation.

I wonder if there's no better way to escape those characters, rather than assume the seasonName will related to the seasonTitle (when spaces or dashes are encountered).
I tried various possibilities to escape (or do exact matches) using seasonName, but it seems to be a starting match (so `facets[seasonTitle]=2` or `facets[seasonTitle]=201` return the same thing).

mediaminister: I found a way to maximise the use of `seasonName`.

resources/lib/vrtplayer/vrtapihelper.py
```diff
@@ -160,19 +160,27 @@ def _map_to_episode_items(self, episodes, titletype=None):
...
+        # Reverse sort seasons if program_type is 'reeksaflopend' or 'daily'
+        seasons = sorted(seasons, key=lambda k: k['key'], reverse=True) if program_type in ('reeksaflopend', 'daily') else seasons
+        for season in seasons:
+            season_key = season.get('key')
+            season_episode_count = str(season.get('doc_count'))
+            title = ''.join((self._kodi_wrapper.get_localized_string(32094), ' ', season_key))
+            # API workaround: season_key matches seasonTitle facet, but seasonTitle behaves as a starting match regex
+            # and therefore sometimes mixes up episodes from different seasons, instead we prefer to use seasonName facet
+            # which behaves as an exact match regex and in most cases matches season_key when spaces are replaced by dashes.
+            if len(episode.get('seasonTitle')) > 1 and episode.get('seasonTitle').replace(' ', '-') == episode.get('seasonName'):
```
dagwieers: Not sure if the `len > 1` is a good identifier. Can we not do i.e. `int()` instead ?

resources/lib/vrtplayer/vrtapihelper.py
dagwieers: While testing various possible syntaxis I got this error:
```
TypeError: facet_value.indexOf is not a function
   at module.exports (/var/app/current/lib/queryBuilders/vrt-video/search.js:83:37)
   at Object.makeESQuery (/var/app/current/lib/queryBuilders/index.js:6:20)
   at handler (/var/app/current/lib/requestHandlers/search.js:21:42)
   at Layer.handle [as handle_request] (/var/app/current/node_modules/express/lib/router/layer.js:95:5)
   at trim_prefix (/var/app/current/node_modules/express/lib/router/index.js:312:13)
   at /var/app/current/node_modules/express/lib/router/index.js:280:7
   at Function.process_params (/var/app/current/node_modules/express/lib/router/index.js:330:12)
   at next (/var/app/current/node_modules/express/lib/router/index.js:271:10)
   at cors (/var/app/current/node_modules/cors/lib/index.js:179:7)
   at /var/app/current/node_modules/cors/lib/index.js:229:17
```
trying: https://vrtnu-api.vrt.be/search?i=video&facets[programUrl]=//www.vrt.be/vrtnu/a-z/universiteit-van-vlaanderen/&facets=seasonTitle:201
This gives a little insight into what was used for the API frontend, i.e. nodejs and queryBuilders

dagwieers:
I like this new implementation, using more than just the "bucket".

resources/lib/vrtplayer/vrtapihelper.py
mediaminister: Not sure what you mean. `seasonTitle` is a string.

dagwieers: I get that, but I assumed that you do `len(episode.get('seasonTitle')) > 1` to filter out integer values as seasonTitle. So I wonder if turning it into an integer wouldn't be a better check.

mediaminister: No, not necessary, I found out that just checking the season data of one episode is not failsafe when VRT uses a different `seasonName` naming convention for the other seasons.
This was the case with "Down the road" https://vrtnu-api.vrt.be/search?i=video&size=150&facets[programName]=down-the-road
So, I tried to filter out as much values that are not affected by the "starting match regex problem" and luckily this fixes the "Down the road" problem.

dagwieers: I don't understand what this has to do with a length of more than 1 though. So I have a big problem if we create conditional statements without a clue to what it relates to. It makes it impossible to refactor in the future not knowing if you break something that worked before.

I assumed you wanted to filter out when seasonTitle was 1-9, now I don't know why you do this. A season could be 10 or 11 too.

dagwieers: From the subtitle you can see the VRT is using the seasonTitle and not the seasonName, so I guess we need to do the same. Somehow I think the solution is in the displayOptions.

PS I discovered it's not actually a "starting match regex", for `2018-2019` the following match: `2018`, `201`, `2`, but also `2019`. What doesn't work is `19` or `18`. A lot of special characters seems to terminate the string (including `-`, `+`, `*` or `|`) although if any special character is repeated before the [A-z0-9] string, these characters are ignored. A very weird implementation...

mediaminister: Values with a length of one are immune for the "starting match regex problem". I filter them out so they are handled by the `seasonTitle` facet.

I don't have a better solution right now to fix "Down the road" season listing:  https://vrtnu-api.vrt.be/search?i=video&size=150&facets[programName]=down-the-road
Down the road has two seasons with `seasonTitle` values: `1` and `2` and `seasonName` values: `2018` and `2`
From `season.get('key')` we get  `1` and `2`  as input values.

So, the most logical thing is to use `seasonTitle` facet here.

dagwieers: Well, that is why I suggested to check if they are integers instead, or maybe even check if the value is `[A-z0-9]+`.

mediaminister: I don't understand, for instance value `1` and value `13` are affected by the "starting match regex problem"

dagwieers: Ok, so my concern here is that if we *have to* implement workarounds for very specific corner-cases, we need to clearly identify these corner-cases in comments (with actual examples) or we have no chance in the future to refactor/improve this code ever. (Ideally we would have proper unit-tests for this at some point).

Now that I understand a bit better why this is more complex, let me suggest some changes in this PR.

PS I'd rather avoid these complexities, but it's clear we have no clue how to escape special characters (which seem to terminate the string matching) at this time, if this is at all possible :-/

mediaminister: Feel free to fix this better. I don't have time this week to look at this again.

dagwieers: The "better" fix will mostly be documenting the corner-cases, but I need to test them myself first, and that won't happen this week...

mediaminister: It seems seasonTitle facet is implemented in the Elasticsearch backend as a partial match. This is not fixable in a good way using api requests. I fixed this again in a different way.

dagwieers:
I made a few cosmetic changes to limit the indentation or improve readability.
Thanks for these fixes/improvements !

PS @pietje666 I think we need a v1.7.2 release with all the stuff we improved recently.

Issue #130, opened 2019-04-13
LiveTV from abroad raises StopIteration exception


dagwieers:
### Describe the bug ###
Starting LiveTV using addon v1.7.1 from NL returns in the below log output:
```
2019-04-14 00:01:18.466 T:53120  NOTICE: [plugin.video.vrt.nu] video_url https://www.vrt.be/vrtnu/kanalen/een/
2019-04-14 00:01:18.815 T:53120  NOTICE: [plugin.video.vrt.nu] Cached token deleted
2019-04-14 00:01:19.378 T:53120  NOTICE: [plugin.video.vrt.nu] Invalid location; required BE
2019-04-14 00:01:20.629 T:53120   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                             - NOTE: IGNORING THIS CAN LEAD TO MEMORY LEAKS!
                                            Error Type: <type 'exceptions.StopIteration'>
                                            Traceback (most recent call last):
                                              File "C:\Users\dwieers\AppData\Roaming\Kodi\addons\plugin.video.vrt.nu\addon.py", line 58, in <module>
                                                router(sys.argv[2][1:])
                                              File "C:\Users\dwieers\AppData\Roaming\Kodi\addons\plugin.video.vrt.nu\addon.py", line 52, in router
                                                vrt_player.play(video)
                                              File "C:\Users\dwieers\AppData\Roaming\Kodi\addons\plugin.video.vrt.nu\resources\lib\vrtplayer\vrtplayer.py", line 65, in play
                                                stream = self._stream_service.get_stream(video)
                                              File "C:\Users\dwieers\AppData\Roaming\Kodi\addons\plugin.video.vrt.nu\resources\lib\vrtplayer\streamservice.py", line 136, in get_stream
                                                roaming_xvrttoken = self.token_resolver.get_xvrttoken(True)
                                              File "C:\Users\dwieers\AppData\Roaming\Kodi\addons\plugin.video.vrt.nu\resources\lib\vrtplayer\tokenresolver.py", line 54, in get_xvrttoken
                                                token = self._get_new_xvrttoken(token_path, get_roaming_token)
                                              File "C:\Users\dwieers\AppData\Roaming\Kodi\addons\plugin.video.vrt.nu\resources\lib\vrtplayer\tokenresolver.py", line 101, in _get_new_xvrttoken
[...]

PR #132, opened 2019-04-14
fixed roaming

mediaminister:
This fixes https://github.com/pietje666/plugin.video.vrt.nu/issues/130


dagwieers:
I can confirm this works now from NL, thanks !

PR #137, opened 2019-04-18
Improve episode listing and sort order

### LINKED ISSUE #126: Other season listing quirks
### Describe the bug ###
We probably don't want to fix any or all of these, but I just wanted to list them here:
- Season listings are not sorted correctly
   - ~Season listing of "Winteruur" shows Season 1, 2, 4 and then 3~ #125
- Season episodes do not always show correctly
  - Episode listing of "Winteruur" for Season 1, 2 and 3 show the same broadcast date
    - Not sure if we have to add the date, now that it is being shown as metadata ?
    - Also not sure if the order makes any sense for these listings either...
  - Episode listing of "Wereldbeeld" for Season 2017 show duplicate episodes (14, 16)
    - Likely an issue at VRT
- The program "Thomas speelt het hard" has an Episode 0 which seems similar to Episode 6
    - Likely an issue at VRT
- ~Episode listing of program "Mag ik u kussen" Season 4 is empty~
- Season listings do not show the program title, which may be confusing
- Episode listing do not show the program-title nor the season
- Also wonder if it would not be better to add Season and Episode in the name
  - So instead of "01/04 - Foo bar" would "S02E03 - Foo bar" not be better ?
    (Only if we find sensible season/episode information)
  - And list all episodes together, rather than have sub-listings per season
    (Would be faster for users)

dagwieers:
This implements more sensible episode names and sort order based on program displayOptions.

This PR includes:
- ~Implementation of *showSeason* (so don't show seasons if this is set to False)~
- Using *broadcast date* when *showBroadcastDate* was set
- ~Using *S0XE0Y* format when *showSeason* is disabled~
- Add sorting methods in GUI listing options
- ~Implementation of sort order (ascending/descending) in GUI based on programType *reeksaflopend*/*reeksoplopend*~
- Implementation of sort by *date added* for recent items
- Implementation for programType *daily* and *oneoff*
- Cleanup related to sort method's.

This fixes #126 as good as can be.

### DIFF
--- a/addon.py
+++ b/addon.py
@@ -33,7 +33,7 @@ def router(params_string):
     params = dict(parse_qsl(params_string))
     action = params.get('action')
     if action == actions.LISTING_AZ_TVSHOWS:
-        vrt_player.show_tvshow_menu_items('az')
+        vrt_player.show_tvshow_menu_items(path=None)
     elif action == actions.LISTING_CATEGORIES:
         vrt_player.show_category_menu_items()
     elif action == actions.LISTING_LIVE:
@@ -41,7 +41,7 @@ def router(params_string):
     elif action == actions.LISTING_EPISODES:
         vrt_player.show_episodes(params.get('video_url'))
     elif action == actions.LISTING_CATEGORY_TVSHOWS:
-        vrt_player.show_tvshow_menu_items(params.get('video_url'))
+        vrt_player.show_tvshow_menu_items(path=params.get('video_url'))
     elif action == actions.PLAY:
         video = dict(
             video_url=params.get('video_url'),
--- a/resources/lib/kodiwrappers/kodiwrapper.py
+++ b/resources/lib/kodiwrappers/kodiwrapper.py
@@ -6,7 +6,6 @@ from __future__ import absolute_import, division, unicode_literals
 import json
 
 import inputstreamhelper
-from resources.lib.kodiwrappers import sortmethod
 import xbmc
 import xbmcgui
 import xbmcplugin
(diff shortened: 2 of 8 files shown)


dagwieers:
@mediaminister When using the episode sorting, it appears as if all the episode values are 0 (they are shown in the listing in this case). Also I have not found a way to reverse the sort order for *reeksaflopend*.

mediaminister:
I found out `displayOptions` is not reliable.
For instance 'Thuis' or 'Down the road' definitely should have a season menu, because there are different seasons with different storylines and characters. This is also the case on VRT NU website:
https://www.vrt.be/vrtnu/a-z/thuis.relevant/
https://www.vrt.be/vrtnu/a-z/down-the-road.relevant/

I'm not sure 'ascending' works reliable in Kodi, I thought sort order is just a GUI toggle. I did use the python `sorted` function on purpose.

dagwieers:
> I found out displayOptions is not reliable.
> For instance 'Thuis' or 'Down the road' definitely should have a season menu, because there are different seasons with different storylines and characters. This is also the case on VRT NU website:
https://www.vrt.be/vrtnu/a-z/thuis.relevant/
https://www.vrt.be/vrtnu/a-z/down-the-road.relevant/

Hmm, I only added `showSeason` support at the very end of this experiment, did not notice that both Thuis and Down the road are problematic. Let's look for a solution :-)

There is at least one issue that I have seen which is that episodes in the same program have no consistent `displayOptions` for some programs. Which causes different titles in the same listing. My solution for this will be to take the more recent episode as the basis for the displayOptions, rather than the episode's displayOptions.

> I'm not sure 'ascending' works reliable in Kodi, I thought sort order is just a GUI toggle. I did use the python sorted function on purpose.
The problem in my testing is mostly that as soon as you change the sort method or sort order in the GUI, it remembers your choice. So for testing it is horrible (and may give you the impression it is not reliable). I still need to find a way to reset those choices.

mediaminister:
Sorting in this pull request is still broken.
I don't like scrolling down to watch the last episode of a daily series.

dagwieers:
Yes, it is not working. It seems impossible to set ascending/descending automatically. Another broken design in Kodi's (python) plugin interface IMO.

dagwieers:
My plan is to restore the old code for everything that is descending (so we control the order) and use the new code for everything that is ascending, so it works as expected for end-users, and they can still manipulate the order as they please.

The only big issue here is that for sorting by episode, it numbers the episodes automatically, which it doesn't when we order it ourselves (and use "unsorted" order). The user can still order by what they like after this (episode or date) but then the episode is visible twice (as a prefix, and in the title). Not a big issue, just a pity.

My other problem here is that for some programs/series the order is descending, while for most programs/episodes it is ascending. So yeah, another disaster in the execution at VRT.NU IMO.

dagwieers:
PS Also not convinced that adding the episode count is of any value, would have been nice if this was also metadata that could be made visible in the skin instead.

PR #138, opened 2019-04-22
Add permalink to video plot

dagwieers:
There's no metadata in the video_dict for external URLs (there really should be though).

So I am adding the shortened permalink to the end of the plot, this could be useful if this episode/video needs to be played through another means (e.g. because Kodi has an issue or the addon fails somehow).

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -16,6 +16,7 @@ disable=
     too-many-branches,
     too-many-instance-attributes,
     too-many-locals,
+    too-many-statements,
     unnecessary-lambda,
     unused-argument,
     wrong-import-order,
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -32,12 +32,16 @@ msgid "Delete cookies"
 msgstr "Delete cookies"
 
 msgctxt "#32010"
-msgid "Subtitles"
-msgstr "Subtitles"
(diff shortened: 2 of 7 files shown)


mediaminister:
The hyperlink isn't clickable. So, this is useless.

dagwieers:
It is not useless, I am using this already.

The URL is short so you can type it in in another mediaplayer or browser to go straight to this episode. It would have been better if we could have added it elsewhere (custom infoLabels) but since there's no other way to share it, the plot will do. Maybe in the future we can improve Kodi to have it visible as proper metadata.

pietje666:
In my opinion i doubt it will be used alot. If users want to see through a browser they would just use the vrt.nu webpage. But as you said you use it already you might 'hide' the functionality through an addon setting. And by default it should be hidden. I guess this is something that will make everyone happy?

dagwieers:
Sure, I usually like to keep the number of settings to a minimum, and I assumed that link would not harm anyone. But if you prefer to not see that link in the plot, that's fine.

It is a pity that we cannot put this anywhere else but the plot though, because it really belongs in the metadata window.

dagwieers:
For this setting I renamed the existing *Subtitles* section into *Interface*, and add the *Show episode permalink in plot* option.

Please review !

mediaminister:
Looks good to me.

PR #139, opened 2019-04-22
Add 'live' to live TV channels

dagwieers:
Ensure 'live' is in the menu entry name so that when these entries are
added to favourites, they clearly indicate this brings you to the live
channels.

At some point we will add listings per channel, so the distinction is
worthwhile in the future (the sooner the better for new users).


mediaminister:
Okay.

PR #140, opened 2019-04-24
Restore the "Default' sort order

dagwieers:
Using SORT_METHOD_UNSORTED is better than using SORT_METHOD_NONE as it appears in the sort order list as 'Default'. So this makes it a bit more consistent, people can go back to the original/preferred sort order.

Also comment out non-functional code.

I removed sort order "Date" as it doesn't work as people would expect, and is confusing as we also have "Date added". The only thing remaining is to make sort order "Name" and sort order "Title" do the right thing (Title should sort discarding the date/episode prefix we added ourselves).

### DIFF
--- a/resources/lib/kodiwrappers/kodiwrapper.py
+++ b/resources/lib/kodiwrappers/kodiwrapper.py
@@ -17,13 +17,14 @@ except ImportError:
     from urllib import urlencode
 
 sort_methods = {
-    'date': xbmcplugin.SORT_METHOD_DATE,
+    # 'date': xbmcplugin.SORT_METHOD_DATE,
     'dateadded': xbmcplugin.SORT_METHOD_DATEADDED,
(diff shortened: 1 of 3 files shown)


dagwieers:
Mental note: Removing all saved plugin states for VRT.NU plugin:
```bash
sqlite3 ~/.kodi/userdata/Database/ViewModes6.db
sqlite> DELETE FROM view WHERE path LIKE '%plugin.video.vrt.nu%';
```

dagwieers:
I think this fixes everything we possibly can.
If anything, our effort should be spend to fix Kodi instead.

dagwieers:
There is a known bug in Kodi that when you select the *Default* sort method, it does not refresh the menu listing. It requires you to leave the menu, and come back to see the *Default* sort method in action.

For other sort methods this works fine, so it is quite confusing when it appears not to work for *Default*.

PR #142, opened 2019-04-24
Add a TV guide menu entry

dagwieers:
So the approach is this:
- You select the day (today, yesterday, ...)
  - We can go back 1 month, and forward 1 week
  - Giving a long list of 40 entries is not a good interface though
- You select the channel
- You get a TV listing for the whole day, with links to play available episodes from the TV-guide

TODO:
- Translate some strings, clean up the interface somewhat

### DIFF
--- a/resources/lib/vrtplayer/tvguide.py
+++ b/resources/lib/vrtplayer/tvguide.py
@@ -25,10 +25,12 @@ CHANNELS = dict(
     ),
 )
 
-DATES = {
-    '1': 'Tomorrow',
-    '0': 'Today',
-    '-1': 'Yesterday',
+DATE_STRINGS = {
+    '-2': 32330,  # 2 days ago
+    '-1': 32331,  # Yesterday
+    '0': 32332,  # Today
+    '1': 32333,  # Tomorrow
+    '2': 32334,  # In 2 days
 }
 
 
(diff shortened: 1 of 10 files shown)


resources/lib/vrtplayer/tvguide.py
```diff
@@ -0,0 +1,122 @@
...
+                        url_dict=dict(action=actions.LISTING_TVGUIDE, date=date, channel=channel),
+                        is_playable=False,
+                        art_dict=dict(thumb=self.__get_media(channel + '.png'), icon='DefaultAddonPVRClient.png', fanart='DefaultAddonPVRClient.png'),
+                        video_dict=dict(plot='TV guide for channel %s' % CHANNELS[channel]['name']),
+                    ),
+                )
+            self._kodi_wrapper.show_listing(channel_items, content_type='files')
+
+        else:
+            dateobj = datetime.strptime(date, '%Y-%m-%d')
```
mediaminister: This causes a crash:
```
dateobj = datetime.strptime(date, '%Y-%m-%d')
TypeError: attribute of type 'NoneType' is not callable
```
It's a known Python bug: https://forum.kodi.tv/showthread.php?tid=112916&pid=1212333#pid1212333

dagwieers: Oh right, coincidentally I was thinking we should document this better in our code, not realizing I was introducing the same thing some hours later :-(

Let's make this a function call instead...

dagwieers:
Another question, I would like to have in Dutch "overmorgen, morgen, vandaag, gisteren en eergisteren", but in English we don't have "overmorgen" or "eergisteren", possibly other languages either. So I was wondering how we would cover this. Leave it empty translated, and cover this in the code that if it is empty in the current language, it isn't used ?

I also noticed that in a Dutch interface, the date-strings are in English nevertheless, so that needs fixing too. And then I still need to look into other hardcoded strings/formatting and how to present it better.

I also would like to indicate in the TV Guide the current time/current playing video, so people can more easily start the current playing episode from the beginning. Any ideas on how to improve the TV Guide are very welcome, as it does not feel completely as it should be.

mediaminister:
Writing code for this "language" problem, seems a bit silly. If there is no single word for it, just use multiple words. You can use "_two days ago_" for eergisteren and "_in two days_" for "overmorgen".

Not sure how to improve the TV guide, maybe we need other GUI methods.
Centering the scrollbar position on the current time item is just an idea, but I don't know it's possible. It will take time to investigate how to make better use of the Kodi GUI.

resources/lib/vrtplayer/tvguide.py
dagwieers: Fixed.

dagwieers:
Seems to work fine for me, the few improvements I have in my head can be added in the future as specific improvements.

PR #144, opened 2019-04-25
only get widevine license url when getting widevine drm streams

mediaminister:
Little improvement: We only need the widevine license url, when getting a widevine drm stream


dagwieers:
Perfect, thanks !

PR #145, opened 2019-04-26
Highlight currently playing program using dateutil

dagwieers:
This PR includes:
- Highlighting the currently playing program (in yellow/red, with *Now on live TV*)
- Introduce dateutil for parsing date/time strings, offers reliability
- Remove the recently introduced **strptime()** function call

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -7,6 +7,7 @@
     <requires>
         <!-- import addon="inputstream.adaptive" version="2.0.19"/ -->
         <import addon="script.module.beautifulsoup4" version="4.5.3"/>
+        <import addon="script.module.dateutil" version="2.6.0"/>
         <import addon="script.module.inputstreamhelper" version="0.3.3"/>
         <import addon="script.module.pysocks" version="1.6.8" optional="true"/>
         <import addon="script.module.requests" version="2.19.1"/>
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 bs4
+python-dateutil
 pylint
 requests
 tox-travis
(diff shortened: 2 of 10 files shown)


dagwieers:
@mediaminister I am hesitant about the existing time-handling in the current code, as it does not include timezone-handling. The current parser is timezone-aware, but we need to provide timezone information if it is not already included in the input.

E.g. the EPG information does include timezone information, but e.g. for the tokenresolver we export UTC time, which IMO is incorrect (and may have an adverse effect). Not sure if this is mandatory, or just out of convenience. Some for other time-related parts. It may affect people in different timezones more than us.

dagwieers:
This is what I was talking about, so the timestamp I am using represents 19:40 CET.
```
$ date --date '@1556044800000'
za feb  4 19:40:00 CET 51279
$ date --date '@1556044800000' --utc
za feb  4 18:40:00 UTC 51279
```

However what we are doing for the cookie handling is this:
```
$ python -c "from datetime import datetime; print datetime.fromtimestamp(1556044800).strftime('%Y-%m-%dT%H:%M:%S.%fZ')"
2019-04-23T20:40:00.000000Z
```

While I think we should be doing this to output UTC ISO 8601:
```
[dag@moria plugin.video.vrt.nu]$ python -c "from datetime import datetime; print datetime.utcfromtimestamp(1556044800).strftime('%Y-%m-%dT%H:%M:%S.%fZ')"
2019-04-23T18:40:00.000000Z
```

dagwieers:
Since the new urllib3 update broke our plugin, I am going to merge this and test everything together before releasing.

mediaminister:
Indeed, time handling for `X-VRT-Token` was wrong. But a cached `X-VRT-Token` expires after one year, so this bug was still unnoticed.
For the short living ` vrtPlayerToken` there was no issue, because the input from https://media-services-public.vrt.be/vualto-video-aggregator-web/rest/external/v1/tokens was already UTC ISO 8601

dagwieers:
One year 😀

PR #152, opened 2019-04-27
Various improvements and fixes

dagwieers:
This PR includes:
- Improvements to regexp for formatted text conversion
- Add localized time strings when using strftime()
- Disable menu caching for certain menus


mediaminister:
Tested this and it's okay for me.

dagwieers:
The localized timestrings only work for platforms that have the support, i.e. LibreELEC does not support anything other than C for the OS, so Python is affected by this as well, and the dates are in English :-(. I'd rather do what Kodi does, but I don't think they expose their internal functions (unfortunately again).

PR #158, opened 2019-04-28
Allow to disable VRT NU credentials

dagwieers:
Probably not worth it.

My idea was to disable all functionality that required credentials (i.e. giving up anonymity).

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -20,20 +20,24 @@ msgid "Credentials"
 msgstr "Credentials"
 
 msgctxt "#32001"
+msgid "Use VRT.NU credentials"
+msgstr "Use VRT.NU credentials"
+
+msgctxt "#32002"
 msgid "Email address"
 msgstr "Email address"
 
-msgctxt "#32002"
+msgctxt "#32003"
 msgid "Password"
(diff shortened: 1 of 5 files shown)


pietje666:
What can people do when they do not use credentials? And i do wonder if there is a demand for that?

mediaminister:
You can watch the livestreams, that's all.
This pull request is useless. There is no real demand for this, in fact we are all forced to create login credentials to use VRT NU.

dagwieers:
Well, in this case only the live streams could work without credentials.
But other content like Radio and Youtube channels could work as well.

mediaminister:
We definitely don't need a user setting for something that works perfectly without.

dagwieers:
The main reason for contemplating this was the need to add something in the first settings-screen that doesn't catch the cursor. Currently if you go into settings, you can't go back to the tabs easily because your cursor is caught in the text-form. With a keyboard or remote control it sucks.

So the only thing we can do is add something before the text input. A slide to turn credentials on/off made the most sense, since another annoying issue is the fact that you end up in the settings/credentials if it was set to empty (even if this was on purpose).

So the slide formalizes the wish to not enter credentials. Yet, you can still browse the catalogue, TV guide or play live streams... Catching 2 birds with one stone.

Issue #57, opened 2019-01-20
Use VRT NU API instead of scraping


pietje666:
As the title says, it would be alot nicer to use an API instead of scraping from the VRT.NU site. 
It would make the addon more stable and most likely be faster.

The problem is, we (or atleast i) do not know which API, nor how to call it.
So any help is welcome!

mediaminister:
I did a little research.

I found out VRT uses some kind of [Elasticsearch](https://www.elastic.co/). I succeeded to get an error message that shows some info about the backend
https://search.vrt.be/search?facets[programUrl]=test&facets[programUrl]=test

I'm not sure which Elasticsearch version or variant VRT uses. I looked at the [Elasticsearch Uri Search documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-uri-request.html) and found out that some parameters work but others don't.

There are two endpoints, a suggest api to get short program info:
https://search.vrt.be/suggest?q=waes
and a search api to get all program info
https://search.vrt.be/search?q=waes

I rewrote the show_videos function using this api as an example:
https://github.com/mediaminister/plugin.video.vrt.nu/blob/c4ff617335ddb2322fcc9c1c7f1d57171166153a/plugin.video.vrt.nu/resources/lib/vrtplayer/vrtplayer.py#L95

The possibilities with this api seem endless, but I have to examene this further.

mediaminister:
I get 241 items with the following request:
https://search.vrt.be/suggest?facets[transcodingStatus]=AVAILABLE

I think this corresponds to the a-z list.

I have not yet found a good query for the categories list.

pietje666:
Looks good for A-Z

pietje666:
for the categories tis seems to be working, i guess the categories are hardcoded => https://search.vrt.be/search?size=150&facets[categories]=docu needs to be mixed with the available facet though

pietje666:
i want to do something this https://search.vrt.be/search?size=150&facets[categories]=docu&select=program&distinct i think it must be possible in some way.

mediaminister:
I tried a lot of different things but did not find a method to get a list of categories. I think this public api is limited. Take a look at this document: https://www.vrt.be/etc/clientlibs/components/search.js

I publish my experiments with this api on: https://github.com/mediaminister/plugin.video.vrt.nu/tree/api

mediaminister:
[...]

PR #161, opened 2019-04-30
Implement fallback for Categories web-scraping

dagwieers:
This PR includes:
- Provide better Studio names
- Fall back to hardcoded list of CATEGORIES if web-scraping fails
- A hardcoded list of channels (both tv as radio)
- Clean up of codebase to use both CHANNELS and CATEGORIES
- Generate list of live streams from CHANNELS
- Include unit test to ensure hard-coded CATEGORIES is identical to web-scraped list

We could now run our tests on an hourly basis to confirm that category-listing scraping works and our internal representation is correct and complete.

This implements https://github.com/pietje666/plugin.video.vrt.nu/issues/57#issuecomment-483650787

PS The fallback mechanism does not support thumbnails. We could do that, but I expect the thumbnails to change more often.

### DIFF
--- a/addon.py
+++ b/addon.py
@@ -25,9 +25,7 @@ def router(params_string):
     addon = xbmcaddon.Addon()
     kodi_wrapper = kodiwrapper.KodiWrapper(_ADDON_HANDLE, _ADDON_URL, addon)
     token_resolver = tokenresolver.TokenResolver(kodi_wrapper)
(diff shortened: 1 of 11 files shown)


dagwieers:
@mediaminister @pietje666 Interested to hear opinions about this one.

dagwieers:
I discovered the unit tests were not actually testing anything (except a function call) as the used examples were invalid.

mediaminister:
Very nice improvement. Okay to merge.

PR #162, opened 2019-04-30
Add Radio support

dagwieers:
As a work-in-progress, I also added Radio stream support as a separate Music addon. I know we already discussed we did not want this in this addon, and it is quite easy to separate this from the other code into a completely different project if needed. (Probably because we want to do this under a different name?)


dagwieers:
Ideas for the future:
- Stream selection (based on preference)
- Working song/artist support for streams other than MP3
- Artwork/fanart support for playing song

pietje666:
Yes indeed a separate addon would be alot better imo

dagwieers:
It currently acts as 2 addons, have you tried it?

pietje666:
No i did not, was looking into it now how do is pass content type as a param? because it still uses the same addon.py right and same logo etc? Did not look into detail

mediaminister:
I'm not against this when VRT integrates audio into VRT NU. But VRT NU is a video service.
For now I think it's better to create a separate project.

dagwieers:
Ok, I was torn. We already include non VRT NU services (like the live streams) and the Radio streams are exactly like the live streams. So if we wanted to do this we may have wanted to rename the addon to just "VRT" instead.

But having both the same addon for Video and Music with the same name and logo could be confusing. If we were going to do this, I would have probably added shortcuts from TV to Radio and back. (Something we could still do from the Settings, so people can easily install one from the other)

dagwieers:
So welcome to **plugin.audio.vrt.radio** :-)

PR #166, opened 2019-04-30
Add Studio Brussel and MNM live streams

dagwieers:
These two streams were still on my wishlist.

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -163,6 +163,22 @@ msgctxt "#32132"
 msgid "Watch Sporza live TV stream (via Internet)"
 msgstr "Watch Sporza live TV stream via (Internet)"
 
+msgctxt "#32141"
+msgid "Studio Brussel live"
+msgstr "Studio Brussel live"
+
+msgctxt "#32142"
+msgid "Watch Studio Brussel live TV stream (via Internet)"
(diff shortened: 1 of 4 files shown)


mediaminister:
This doesn't work in Kodi 17, because you hardcoded the mpeg_dash streams that are only supported by inputstream.adaptive.

I have a fix for this, you can just provide the livestream `video_id` and let the streamservice select the right stream.
https://github.com/mediaminister/plugin.video.vrt.nu/commit/1642cc17806823d4e8b66110c610dcd79640a713

dagwieers:
@mediaminister I first looked at scraping the website, but that was a big pile of Javascript junk. I see this as a potential way to fallback from scraping for the VRT TV channels as well. If scraping fails, we could use the last known working link. It still depends on how often that link would change.

mediaminister:
We can use the livestream `video_id` for all channels as a fallback.
But I think using mpeg_dash and hls urls is not a good idea because these urls have changed a couple of times already.

dagwieers:
@mediaminister Perfect, I am reworking your PR with this in mind. The death to web scraping dependencies !

I wonder whether we need to inform the user when we fall back when web-scraping fails, or whether we just rely on unit tests to inform us?

mediaminister:
It's not nice to bother a "normal" user with this. The user only cares if the stream works. Not why it works.

PR #167, opened 2019-05-01
livestream compatibility for kodi 17

dagwieers:
This is the fix from @mediaminister for Kodi 17 support.


dagwieers:
I cannot seem to push to this branch, so will create #168 instead.

PR #168, opened 2019-05-01
Fallback for live streams when web scraping fails

### LINKED ISSUE #57: Use VRT NU API instead of scraping

dagwieers:
This implements vualto live stream support (by video_id) for streams
that do not rely on web scraping, but also implements a fallback
mechanism in case web scraping fails.

This is the last piece to get rid of any reliance on web scraping.
Users everywhere rejoice !

This fixes #57

### DIFF
--- a/resources/lib/vrtplayer/__init__.py
+++ b/resources/lib/vrtplayer/__init__.py
@@ -32,7 +32,7 @@ CHANNELS = {
         name='Eén',
         studio='Een',
         live_stream='https://www.vrt.be/vrtnu/kanalen/een/',
-
+        live_stream_id='vualto_een',
     ),
     'canvas': dict(
         id='1H',
@@ -40,6 +40,7 @@ CHANNELS = {
         name='Canvas',
         studio='Canvas',
         live_stream='https://www.vrt.be/vrtnu/kanalen/canvas/',
+        live_stream_id='vualto_canvas',
     ),
     'ketnet': dict(
         id='O9',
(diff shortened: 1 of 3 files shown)


dagwieers:
@mediaminister Please test!

I tested this by enabling (non-working) stubru and mnm livestreams. And it correctly falls back to using the provided video_id.

dagwieers:
I am going to merge this, so I can test everything (including Radio support) in a single build.
Feedback and improvements welcome !

mediaminister:
This code can't work. You can't return streamurls before we check roaming, etc...
Please wait

dagwieers:
@mediaminister OK, this affects only the fallback for the original streams. We can fix this.

mediaminister:
Maybe it affects more than that. I'm testing at the moment.

dagwieers:
No, it worked fine (as a fallback) for the MNM and Studio Brussel streams, and only is used as a fallback for the original streams. So yes, we need to fix this, it definitely fails to fallback.

mediaminister:
We don't use scraping for MNM and Studio Brussel, it never executed your fall back code.

dagwieers:
Well, for testing I added the live URL that obviously fails to be scraped correctly.
That is how I tested where and how to inject it.

dagwieers:
I'll be back later this evening.

mediaminister:
I found out what was wrong and created a fix: https://github.com/pietje666/plugin.video.vrt.nu/pull/170

dagwieers:
I wondered what the _geo streams were, apparently they are the solution ;-)

PR #170, opened 2019-05-01
improve streamservice

mediaminister:
`vualto_een`, `vualto_canvas`, `vualto_ketnet` and `vualto_sporza` are deprecated video_id's since VRT geoblocks live streams. This was the main cause the previous commit was failing on hls streams.

Other changes:
- removed direct mpeg_dash option, for two reasons:
    - every VRT stream is available through vualto video aggregator api
    - in the past direct stream urls changed several times a year at unpredictable times
- separated web scraping function, so it's easier and more failsafe to fix
- added an error message to Kodi log when web scraping fails
- centralized `api_data` manipulations in `_get_api_data` function

### DIFF
--- a/resources/lib/vrtplayer/__init__.py
+++ b/resources/lib/vrtplayer/__init__.py
@@ -32,7 +32,7 @@ CHANNELS = {
         name='Eén',
         studio='Een',
         live_stream='https://www.vrt.be/vrtnu/kanalen/een/',
-        live_stream_id='vualto_een',
+        live_stream_id='vualto_een_geo',
     ),
     'canvas': dict(
         id='1H',
@@ -40,7 +40,7 @@ CHANNELS = {
         name='Canvas',
         studio='Canvas',
         live_stream='https://www.vrt.be/vrtnu/kanalen/canvas/',
-        live_stream_id='vualto_canvas',
+        live_stream_id='vualto_canvas_geo',
     ),
     'ketnet': dict(
         id='O9',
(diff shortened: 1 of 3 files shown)


dagwieers:
Thanks, looks good and makes it easier to test the web-scraping code in unit tests.
Which is something we will need to test in an automated fashion soon.

The only thing I do not like is the long try-except block, this is unneeded IMO, if the web-scraping fails, it fails early on. The other code should not be throwing an exception (and if it does, we need to know and handle it accordingly).

mediaminister:
I placed all html data-* attributes in the try-except block, because it's data we scrape from an html page.
We really don't know where on the html page it will fail first. VRT can change attributes while preserving html tags and vice versa.

dagwieers:
Well, the data-attributes won't fail (i.e. not return an exception), they will simply return None.
I am proposing a change.

dagwieers:
Also, I would like to enable the Sporza stream by default, even if it currently only shows black.
(In fact I wanted to test this during the Cup Final this afternoon, but couldn't make it work, now we know why).

mediaminister:
Gives an error:                                                
```
self._kodi_wrapper.log_error('Web scraping api data failed: %s', e)
TypeError: log_error() takes exactly 2 arguments (3 given)
```

dagwieers:
I just found out myself! Great. Everything else looks fine, until it breaks differently ;-)
Let's consider this as good as it gets for now.

PS What is your opinion to Radio support ? I'd like to know whether we can keep it as-is, or start a second project for this.

PR #171, opened 2019-05-02
Add icon support

dagwieers:
This introduces menu item icons.


dagwieers:
@mediaminister @pietje666 Please provide feedback. I'd like to keep the number of open branches to a minimum, especially since I like to work on the new routing code (which works nice for **plugin.audio.vrt.radio**).

resources/lib/kodiwrappers/kodiwrapper.py
```diff
@@ -198,6 +199,12 @@ def get_proxies(self):
     def has_inputstream_adaptive_installed(self):
         return xbmc.getCondVisibility('System.HasAddon("{0}")'.format('inputstream.adaptive')) == 1
 
+    def has_white_icons(self):
+        return xbmc.getCondVisibility('System.HasAddon("resource.images.studios.white")') == 1
+
+    def has_coloured_icons(self):
```
pietje666: so this will work when you manually install the color addon? Otherwise its white?

pietje666: nvm saw the comment

resources/lib/vrtplayer/vrtplayer.py
```diff
@@ -87,14 +87,28 @@ def show_category_menu_items(self):
         self._kodi_wrapper.show_listing(category_items, sort='label', content_type='files')
 
     def show_livestream_items(self):
+
+        if self._kodi_wrapper.has_white_icons():
```
pietje666: people will only need the if when they manually uninstall the white icons addon. I find it a bit unnecessary

resources/lib/kodiwrappers/kodiwrapper.py
dagwieers: @mediaminister also created the coloured icons, but they have been reverted as they were in the wrong format. We'll have to wait for them to be resubmitted and released :-/

resources/lib/vrtplayer/vrtplayer.py
dagwieers: I added it as a dependency to the addon.xml so the conditional is no longer necessary. Unless you do not want to have it as a dependency (just optional).

dagwieers: If you accept the new dependency, I can indeed remove the conditional statements.

PS If we have the coloured icons as well, I'd like to use the white icons for the fanart, and coloured for the thumbnail/icon.

pietje666: To me it looks good feel free to merge already :)

PR #174, opened 2019-05-03
Replace requests with urllib2

mediaminister:


resources/lib/vrtplayer/streamservice.py
```diff
@@ -3,12 +3,17 @@
 # GNU General Public License v3.0 (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 from __future__ import absolute_import, division, unicode_literals
+import json
+
 from resources.lib.helperobjects import apidata, streamurls
 
 try:
-    from urllib.parse import urlencode
+    from urllib.parse import urlencode, quote
```
dagwieers: Could we maybe move this block into its own file, and import from it instead ?
This would allow us to import the urllib-specific only when needed.

dagwieers: Let's do this in a separate PR :-)
Thanks !

dagwieers:
Looking good.

PR #180, opened 2019-05-04
Use different string ids to not corrupt Kodi

dagwieers:
Since there is something corrupting my language strings in Kodi, I started investigating and our plugin is using the wrong range for the string ids. This fixes it.

Not sure if our plugin was the cause though, time will tell.


dagwieers:
This range is what the vast majority of the official plugins are using.

Issue #172, opened 2019-05-02
Lower quality stream/ ISP limits


vasonline:
### Describe the bug ###

Might not be a bug with the addon...

Situation: went over download budget with ISP.
Download speed is now limited to ~100kB/s

Videostream from addon play for 2 sec, then buffers, then plays 2 sec.
Clearly the restricted bandwidth is playing a role here.
! no fault of the addon.

However VRT website does something different. When playing livestream
or using on demand it switches to a lower quality stream.
Playback then functions normal (no more buffer empty) at a lower quality.

I do not know, maybe the addon can implement something similar if it notices
the bandwidth is not sufficient?
A sort of auto fallback?

Greetings:

ps @dagwiers: Your EPG is great. MUCH better to use with the Kodi
remote.

dagwieers:
@vasonline I will put it on the [Ideas](https://github.com/pietje666/plugin.video.vrt.nu/wiki/Ideas) Wiki page, but leave it to others to see if this is feasible (either by auto-sensing or by specifying a bandwidth limit).

mediaminister:
You can also solve this issue immediately by changing ISPs. In my humble opinion, this is the best solution if you use internet streaming to watch VRT on your TV set.

This function is called "[adaptive bitrate switching](https://en.wikipedia.org/wiki/Adaptive_bitrate_streaming)" and this is not yet activated in [inputstream.adaptive](https://github.com/emilsvennesson/script.module.inputstreamhelper), the module Kodi uses for streaming. It will take some time to investigate en write code to get this working properly.

In the coming weeks I'll investigate how to implement a temporary fix with a user setting in our own addon.

dagwieers:
Given the fact how well Kodi works on Android, being able to limit the bandwidth would be useful for using the VRT NU addon on mobile networks too.

vasonline:
Hi,

@mediaminister: interesting read on link!
- Could be benefical too for a faster first load of the stream

@dagwiers:
Yes, mobile could a good reason too.

mediaminister:
I have a first simple fix. When you disable the DRM setting in our add-on and disable the add-on inputstream.adaptive, you can use the Kodi bandwidth setting in System > Internet access > Internet connection bandwidth limitation again.
Pull request: https://github.com/pietje666/plugin.video.vrt.nu/pull/182
Zip:
[...]

PR #182, opened 2019-05-04
allow Kodi to select lower bandwidth hls streams

mediaminister:
A first simple fix for https://github.com/pietje666/plugin.video.vrt.nu/issues/172

### DIFF
--- a/resources/lib/vrtplayer/streamservice.py
+++ b/resources/lib/vrtplayer/streamservice.py
@@ -255,37 +255,42 @@ class StreamService:
 
         return stream_url
 
-    # Speed up HLS selection, workaround for slower kodi selection
     def _select_hls_substreams(self, master_hls_url):
-        import re
-        base_url = master_hls_url.split('.m3u8')[0]
-        # m3u8 = requests.get(master_hls_url, proxies=self._proxies).text
-        m3u8 = urlopen(master_hls_url).read()
+        '''Select HLS substreams to speed up Kodi player start, workaround for slower kodi selection'''
         direct_audio_url = None
         direct_video_url = None
         direct_subtitle_url = None
 
-        # Get audio uri
-        audio_regex = re.compile(r'#EXT-X-MEDIA:TYPE=AUDIO[\w\-=,\.\"\/]+URI=\"([\w\-=]+)\.m3u8\"')
-        match_audio = re.findall(audio_regex, m3u8)
-        if match_audio:
-            direct_audio_url = match_audio[len(match_audio) - 1]
+        bandwidth_limit = self._kodi_wrapper.get_global_setting('network.bandwidth')
 
-        # Get video uri
-        video_regex = re.compile(r'#EXT-X-STREAM-INF:[\w\-=,\.\"]+[\r\n]{1}([\w\-=]+\.m3u8(\?vbegin=[0-9]{10})?(&vend=[0-9]{10})?)[\r\n]{2}')
-        match_video = re.search(video_regex, m3u8)
-        if match_video:
-            direct_video_url = base_url + match_video.group(1)
+        # Only select urls from Unified Streaming Platform and when no bandwidth limit is set in Kodi
+        if any(x in master_hls_url for x in ('.ism/', '.isml/')) and bandwidth_limit == 0:
+            import re
+            base_url = master_hls_url.split('.m3u8')[0]
+            m3u8 = urlopen(master_hls_url).read()
 
-        # Get subtitle uri
-        if self._kodi_wrapper.get_setting('showsubtitles') == 'true':
-            subtitle_regex = re.compile(r'#EXT-X-MEDIA:TYPE=SUBTITLES[\w\-=,\.\"\/]+URI=\"([\w\-=]+)\.m3u8\"')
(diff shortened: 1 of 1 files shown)


dagwieers:
I would prefer to delay this until after the v1.9.0 release (this weekend ?).
And maybe even until it's a bit more full-featured?

If we keep this branch in sync with the master branch, people that need it can use this branch.
This reminds me to look into automatic repository with versions from branches so people can enable this in Kodi for testing. This would be ideal for things like this we want to experiment with. Or to get fixes out to people faster.

mediaminister:
This pull request repairs Kodi's own video quality selection function in case the user has set a bandwidth limitation, but this only works with HLS streams. I'm still investigating but it seems inputstream.adaptive has a similar bandwidth setting.

So, I doubt this will become a full feature in our add-on. A proper fix definitely need adjustments upstream.

dagwieers:
@mediaminister I don't really understand how this PR works. I think I miss some context on how things now work. I am also not sure if this is still useful if #189 works as expected.

mediaminister:
I'll explain. When inputstream.adaptive with Widevine DRM-support is not available (for example in Kodi 17), we need to use the HLS-streams VRT (still) provides. HLS-streams work directly in Kodi player.

But, as Kodi player is relatively slow to select a HLS substream, I wrote the `_select_hls_substreams ` function to always select the highest quality HLS-substream and send that stream directly to Kodi player. This speeds up video playback.
The `_select_hls_substreams ` function also contains a workaround to provide webvtt subtitles for video on demand items, because Kodi player doesn't handle this right.

This pull request basicly adds an` 'if else'` structure to the `_select_hls_substreams ` function so when the user has set a bandwith limit, the function does nothing.

Maybe this pull request isn't usefull anymore, I'll look into this at a later time when I have more time.

You should know that some workarounds are needed in a Kodi 17/ HLS-streaming context:
- webvtt subtitling
- faster video playback

dagwieers:
@mediaminister Ok, that clears it up. One thing I wondered was why we only specify **inputstream.adaptive** for DASH streams and not HLS. But I guess that's because we make that selection ourselves. Could we maybe add some additional notes/comments to the code for this ? Like explicitly mention Kodi 17 for specific workarounds.

So am I right that for Kodi 18 we don't have any specific workarounds ? Or should we report something upstream to get it fixed properly there ?

Also, why don't we set inputstream.adaptive for HLS streams in a Kodi 18 context ?

dagwieers:
It is also clear I need my own Kodi 17 setup for testing...

dagwieers:
> But, as Kodi player is relatively slow to select a HLS substream

How slow is relatively slow ?

I noticed that playing a stream is already relatively slow. I haven't found a way to speed it up. For my VRT Radio plugin I did notice that playing stream directly, rather than another ListItem round-trip does help (but obviously requires Kodi to handle everything correctly from the start and no additional handling required).

mediaminister:
@dagwieers We have to take a lot into account when selecting the streams:
- subtitles, audio and video should always work on any Kodi platform
- enabling DRM is a free user choice
- inputstream.adaptive and/or DRM doesn't work in earlier Kodi versions
- Kodi player doesn't support fragmented WebVTT subtitling in HLS and doesn't support MPEG-DASH at all.

And now we have to take into account that bandwidth is also a free user choice, so it gets more complicated.

Not using inputstream.adaptive for HLS is a choice to keep `streamservice` as simple as possible. It means we use the HLS and subtitling workaround also on Kodi 18. So we limit combinations of al these different factors and make sure subtitling always works. The workarounds are not only related to Kodi versions but also to the DRM user setting or to inputstream.adaptive.

> Could we maybe add some additional notes/comments to the code for this ?
Good idea, I'll do this when I find the time.

> Or should we report something upstream to get it fixed properly there ?
Not sure yet what we need, I'll first look at your fix.

> Also, why don't we set inputstream.adaptive for HLS streams in a Kodi 18 context ?
I didn't test the VRT HLS streams, so I'm unsure if subtitling works well.

> How slow is relatively slow ?
When I tested last year, my substream selection workaround saved one or two seconds.

dagwieers:
> - inputstream.adaptive and/or DRM doesn't work in earlier Kodi versions

At the moment I don't think we support anything other than v17 (Krypton) and v18 (Leia).

> - Kodi player doesn't support fragmented WebVTT subtitling in HLS and doesn't support MPEG-DASH at all.

Ok, but inputstream.adaptive does support HLS and MPEG-DASH, right ? It's only fragmented WebVTT subtitling in HLS that is broken ? If so, let's open an issue upstream and track it. We can then refer to this upstream ticket in the workaround code.

I don't mind the workarounds at all (even if I wasn't aware of it all) but if we can report/fix stuff upstream we can slowly get rid of the extra code over time.

Another thing I would like to fix is that our code includes logging (Info|Verbose|Debug) so that it becomes more clear that when something does not work correctly, we know exactly what the code is doing and the user has a more informed guess at what could be the problem. I already added this to the settings, we now have to expose this to our log-functions and add sufficient entries.

I hope with more information I can determine why I have never seen 720p streams working. And I don't know if it's because they are not available, or I have an issue on my system.

Issue #176, opened 2019-05-03
Restructure code


dagwieers:
### Describe the bug ###
A few thoughts on code structure:
- Kodi's included path for code is the main directory, not `resources/lib`
- We have issues because we include/import too much code that is unneeded
- A lot of functionality is locked inside classes, which complicate their use needlessly
- The current nesting of of classes/methods is too complicated
- Design should enable unit testing (so XBMC stuff should stay separated)

I would like to propose:
- Moving all code into a single directory (e.g; `./vrtnu/`) with potentially 10 to 15 files
- Use **script.module.routing** to improve addon entrance (see #103)
- Get rid of `actions.py` and the unwanted `__init__.py` files
- Separate files on function and reusability
- Import only needed data/classes/methods so it looks much cleaner to what is used
- Separate independent functions instead of static class methods (so they can be imported)
- Simplify TitleItem, or replace it with a well-designed dictionary

I am very skeptical to forced object oriented programming, it's nice where it improves readability, but sometimes putting stuff in classes together because you have to overcomplicates the design (and Python does not force you to).

pietje666:
Maybe some static class methods could be replaced indeed. How do you mock stuff when using procedural  programming.

PR #187, opened 2019-05-05
WIP: Reorganize structure and design

### LINKED ISSUE #103: Move to using script.module.routing
**Describe the bug**
Yesterday I stumbled on **plugin.video.fosdem** and I was charmed by how simple and easy the code was to understand. Now I am convinced we ought to be using **script.module.routing** as well (I added it to the refactoring roadmap).

See https://github.com/jelly/plugin.video.fosdem/blob/master/addon.py#L56

More information related to **script.module.routing**: 
- Documentation: https://github.com/tamland/kodi-plugin-routing
- Tutorial: https://github.com/afrase/kodiswift/blob/master/docs/tutorial.rst

dagwieers:
Things I would like to do:
- Use a routing library for having a more simple interface
- Rename overly long functions, e.g. get_localized_string()
- Move all python files into a single directory

**This is a work-in-progress.**

This fixes #103 and #176


dagwieers:
One big problem with **script.module.routing** is that URLs (or anything with slashes) cannot be passed through. Another problem is that the global **plugin** object needs to be passed on as well, while I had hoped that this would not be needed. So it's an even more radical change than I had hoped.

I have some working code for some of the menu listings, but it requires a lot more work. Feel free to discuss any design changes in the current PR, because there's a lot going on.

mediaminister:
I don't like script.module.routing, I think a separate navigation function is good enough. For instance: https://github.com/mediaminister/plugin.video.vrt.nu/blob/fast/resources/lib/vrtnu/vrtnu.py#L24

dagwieers:
So my approach for this will be to reorganize everything except the existing routing, so that existing paths do not break. However, I think we need to redesign that part at some point as well. But we can go to great length without affecting this for now.

dagwieers:
Some of this will happen in another PR, some day.

PR #188, opened 2019-05-05
Enable 'reuselanguageinvoker'

dagwieers:
This ought to speed up using the addon as it avoids loading the python
interpreter and libraries.

I don't see a noticable speed difference for v1.9.0 though.

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -94,5 +94,6 @@ v1.7.0 (2019-03-26)
             <screenshot>resources/media/screenshot02.jpg</screenshot>
             <screenshot>resources/media/screenshot03.jpg</screenshot>
         </assets>
+        <reuselanguageinvoker>true</reuselanguageinvoker>
     </extension>
 </addon>


dagwieers:
Based on testing using this option with v1.8.0 + latest broken **urllib3** shows that initial load times are +4 seconds, but after that the addon is as fast as our current **urllib2** + selective imports.

Combining **requests** + selective imports brings us all the benefits, fast addon access and ease-of-use **requests** library. The only impact is there on the very first HTTP request, which IMO is acceptable (especially since there is a fix coming for this **urllib3** regression).

My only remaining concern with using **requests** or **urllib3** is the additional dependency and future breakage/regressions like we have experienced lately. I am open to changing my opinion. I have a strong preference for keeping the codebase as clean/simple as possible.

dagwieers:
Going to merge this, as it has no negative impact. IMO this is something all plugins should be doing.

PR #189, opened 2019-05-06
Implement inputstream.adaptive.max_bandwidth

### LINKED ISSUE #172: Lower quality stream/ ISP limits

dagwieers:
Attempt to use max_bandwidth for inputstream.adaptive.

This will select the lowest value of both the global network max_bandwidth as well as our own max_bandwidth setting.

This fixes #172

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -47,6 +47,10 @@ msgctxt "#30021"
 msgid "Use DRM for 720p HD-quality livestreams"
 msgstr "Use DRM for 720p HD-quality livestreams"
 
+msgctxt "#30022"
+msgid "Maximum bandwidth (bits/sec)"
+msgstr "Maximum bandwidth (bits/sec)"
+
 msgctxt "#30030"
 msgid "Advanced"
 msgstr "Advanced"
--- a/resources/language/resource.language.nl_nl/strings.po
+++ b/resources/language/resource.language.nl_nl/strings.po
@@ -48,6 +48,10 @@ msgctxt "#30021"
 msgid "Use DRM for 720p HD-quality livestreams"
 msgstr "Gebruik DRM voor de livestreams in 720p HD-kwaliteit"
 
+msgctxt "#30022"
+msgid "Maximum bandwidth (bits/sec)"
+msgstr "Maximale bandbreedte (bits/sec)"
+
 msgctxt "#30030"
 msgid "Advanced"
 msgstr "Geavanceerd"
--- a/resources/lib/kodiwrappers/kodiwrapper.py
+++ b/resources/lib/kodiwrappers/kodiwrapper.py
@@ -97,6 +97,8 @@ class KodiWrapper:
     def play(self, video):
         import xbmcgui
         play_item = xbmcgui.ListItem(path=video.stream_url)
+        play_item.setProperty('inputstream.adaptive.max_bandwidth', str(self.get_max_bandwidth() * 1000))
+        play_item.setProperty('network.bandwidth', str(self.get_max_bandwidth() * 1000))
         if video.stream_url is not None and video.use_inputstream_adaptive:
             play_item.setProperty('inputstreamaddon', 'inputstream.adaptive')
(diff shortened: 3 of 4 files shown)


dagwieers:
This does not work as I expected. The settings do not show the pull-down menu as the main settings page. And whatever value inserted does not appear to work.

mediaminister:
Thanks, this can work for `mpeg_dash` streams. Just check if inputstream.adaptive is enabled, DRM is enabled and check if you're using integers in the `get_max_bandwidth` function.

dagwieers:
@mediaminister I am not sure if this is related to DRM only. AFAICT it works for **inputstream.adaptive** in general. In fact, I think it's even better to always add this property so it also works if somehow Kodi decides to use **inputstream.adaptive**.

dagwieers:
And I am now also setting **network.bandwidth** for every stream, and it simply works.

So with this PR we successfully limit the bandwidth on all the streams. The only thing not limited is our own HTTP requests, and I am not sure if we really have to. Maybe we ought to provide a drop-down list of values like the global setting.

PS It now no longer belongs to the **DRM** settings tab, so again where do we put this ? I think we have to reduce **Credentials**, **Subtitles**, **DRM**, **Advanced** into **Credentials**, **Interface**, **Video** and then move Subtitles and DRM to **Video**.

mediaminister:
For now, it definitely doesn't work for hls streams, because these streams are handled by `_select_hls_substreams` function which always selects the highest quality stream before sending the url to Kodi player.

dagwieers:
@mediaminister I understand now. Our best hope for the future is to have both HLS and DASH be handled by Kodi itself. In the meantime we have to handle bandwidth limitation ourselves as best as we can for HLS.

So I'll merge both PRs, and change the settings in a follow-up PR.

mediaminister:
> And I am now also setting network.bandwidth for every stream, and it simply works.
@dagwieers Are you sure this really works, for example when you disable DRM and play a livestream?

dagwieers:
@mediaminister I will test. Using **network.bandwidth** did work for a scenario where **inputstream.adaptive.max_bandwidth** did not work though. But it would help if the logs would indicate which scenario I am effectively testing (now I simply try a few things that I assume are different scenarios, but can't know for sure...)

For debugging, we may want to force MPEG-DASH or HSL ourselves, maybe ? Maybe a new option in settings ?

dagwieers:
And maybe we should Mock ListItem so we can actually test if we get the expected ListItem entries for each of the scenarios based on settings/bandwidth.

mediaminister:
Disable the DRM user setting and play Eén or Canvas livestream and you get always HLS. The streaming protocol is already displayed in the log. I don't think we need other settings.

dagwieers:
Wait until you see the implementation I have in my head... :-)
(Not sure if everything is possible though, but would be neat!)

dagwieers:
@mediaminister I can confirm that "network.bandwidth" does not work. So it means for streams played by Kodi/HLS, only the global setting takes effect. So probably the best way forward is to remove the bandwidth setting from our plugin, and only take the global setting into account. Unless we specify that this setting only affects *true* inputstream.adaptive-induced streams.

dagwieers:
I finally found why 720p HD streams were not working. By default InputStream Adaptive has a configuration setting **Ignore Display Resolution** which is disabled. I guess since my display resolution is 1080p, which could not be matched, it did not select 720p.

I updated our documentation, and add a [HD quality](https://github.com/pietje666/plugin.video.vrt.nu/wiki/HD-quality) section.

PR #190, opened 2019-05-06
Reorganize settings

dagwieers:
This PR implements:
- **Credentials**
  - Username
  - Password
- **Video**
  - Show subtitles
  - Use DRM
  - ~Internet bandwidth limitation~
- **Interface**
  - Permalink in plot
- ~**VRT Radio**~
  - ~Install/run VRT Radio~
- **Debug**
  - Inputstream Adaptive settings
  - Clear VRT cookies
  - Log level

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,15 +101,15 @@ you can send a message to [our Facebook page](https://facebook.com/kodivrtnu/).
 - Roaming support added (@mediaminister)
 
 #### v1.5.0 (2018-12-27)
- - 720p Livestreams when enabling in settings + having kodi 18 + having widevine.dll present (@mediaminister)
- - Fixed bug where watched icon was not showing in Kodi 18
- - Implemented different way of working with subtitles (@mediaminister)
+- 720p Livestreams when enabling in settings + having kodi 18 + having widevine.dll present (@mediaminister)
+- Fixed bug where watched icon was not showing in Kodi 18
+- Implemented different way of working with subtitles (@mediaminister)
 
 #### v1.4.3 (2018-11-07)
- - Livestreams working again
+- Livestreams working again
 
 #### v1.4.2 (2018-10-11)
- - Changed way of working with urls when a season is refering to href="#"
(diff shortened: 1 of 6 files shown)


mediaminister:
VRT Radio and Debug is useless, everyone who wants to enable these settings or add-ons, can do this in Kodi.
We don't need to duplicate functionality, keep it simple please.

dagwieers:
@mediaminister The Radio section is to make it easier to find VRT Radio, and I plan to do the same in the VRT Radio plugin. But only if it can both install it (if needed) and run it. (Currently it only runs it).

For **Debug** I do think it is useful to have this readily available. And both Retrospect as well as Netflix plugin are doing the same. The Log level only affects the plugin log settings, which I think is an important property as well. Kodi's debug logging is too verbose.

mediaminister:
1) You can easily find and install other add-ons through Kodi's search function. I thought we agreed to make this a separate project because VRT NU is a video service.
2) Debug is okay for me, but I think most users will never use this the debug option.

dagwieers:
Things I would like to see in the future as well:
- Parental
  - Exclusive channel list (limit everything based on channel list)
  - Day/Time settings (only enable plugin between specific times)
- **Debug**
  - (Re-)install Widevine
  - Clear tokens
  - Force stream selection (MPEG-DASH/HLS/...) *(This may report failure if it is not possible)*

dagwieers:
@mediaminister VRT Radio *is* a separate project, but I think it is in the user's best interest as well as beneficial to both projects to link to each other as they are both complementary. There are other more plugins that do similar things.

@pietje666 What do you think ? If you both object obviously I won't do it, but this is what I would propose.

dagwieers:
In the longer run I was even thinking to link to other plugins that relate to Flemish (or even Belgian) TV stations. Have a **More...** tab with Retrospect Belgium, Yelo, and others. But that would be an even more controversial idea ;-)

PR #196, opened 2019-05-08
Improve bandwidth limitation settings

dagwieers:
Make it clear this only affects inputstream.adaptive.

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -32,16 +32,24 @@ msgid "Video"
 msgstr "Video"
 
 msgctxt "#30011"
+msgid "Subtitles"
+msgstr "Subtitles"
+
+msgctxt "#30012"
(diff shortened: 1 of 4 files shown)


dagwieers:
@mediaminister @pietje666 I merged this quite quickly, but I would still appreciate your review/feedback/criticism. The only reason for me to merge this quickly is to make it easier to merge it with another large feature-branch for testing. I know this is not perfect, but in this case most (if not all) of the changes in this PR are fine.

My only concern is whether we actually want to provide our own bandwidth limitation if it only affects InputStream.Adaptive. I plan to add "network.bandwidth" support to Kodi for ListItem's, but that's a long-term plan.

mediaminister:
It's totally okay for now to merge this, but please don't make a new release yet. I'm rewriting `_select_hls_substreams` so the `max_bandwidth` setting will also work without inputstream.adaptive enabled. I'll add a pull request shortly.

Issue #191, opened 2019-05-06
URLError: <urlopen error ('_ssl.c:710: The handshake operation timed out',)>


dagwieers:
### Describe the bug ###
I just got this error while testing the MNM stream
```
2019-05-06 16:52:02.405 T:1732186992   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                             - NOTE: IGNORING THIS CAN LEAD TO MEMORY LEAKS!
                                            Error Type: <class 'urllib2.URLError'>
                                            Error Contents: <urlopen error ('_ssl.c:710: The handshake operation timed out',)>
                                            Traceback (most recent call last):
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/addon.py", line 62, in <module>
                                                router(sys.argv[2][1:])
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/addon.py", line 56, in router
                                                vrt_player.play(params)
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/resources/lib/vrtplayer/vrtplayer.py", line 184, in play
                                                stream = stream_service.get_stream(params)
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/resources/lib/vrtplayer/streamservice.py", line 193, in get_stream
                                                video_json = self._get_video_json(api_data)
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/resources/lib/vrtplayer/streamservice.py", line 141, in _get_video_json
                                                playertoken = self.token_resolver.get_live_playertoken(token_url, api_data.xvrttoken)
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/resources/lib/vrtplayer/tokenresolver.py", line 58, in get_live_playertoken
                                                token = self._get_new_playertoken(token_path, token_url, headers)
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/resources/lib/vrtplayer/tokenresolver.py", line 79, in _get_new_playertoken
                                                playertoken = json.loads(urlopen(req).read())
                                              File "/usr/lib/python2.7/urllib2.py", line 154, in urlopen
                                              File "/usr/lib/python2.7/urllib2.py", line 429, in open
                                              File "/usr/lib/python2.7/urllib2.py", line 447, in _open
                                              File "/usr/lib/python2.7/urllib2.py", line 407, in _call_chain
[...]

PR #197, opened 2019-05-08
tryfix urllib2 https handshake timeout

mediaminister:
tryfix https://github.com/pietje666/plugin.video.vrt.nu/issues/191

### DIFF
--- a/resources/lib/vrtplayer/tokenresolver.py
+++ b/resources/lib/vrtplayer/tokenresolver.py
@@ -75,7 +75,8 @@ class TokenResolver:
 
     def _get_new_playertoken(self, path, token_url, headers):
         import json
-        req = Request(token_url, data='', headers=headers)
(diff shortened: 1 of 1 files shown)


resources/lib/vrtplayer/tokenresolver.py
```diff
@@ -75,7 +75,8 @@ def get_cookie_from_cookiejar(cookiename, cookiejar):
 
     def _get_new_playertoken(self, path, token_url, headers):
         import json
-        req = Request(token_url, data='', headers=headers)
+        req = Request(token_url, headers=headers)
+        req.get_method = lambda: 'POST'
```
dagwieers: No offence, but that looks... strange. Why the lambda function here ?

mediaminister: Because `get_method` is a function and a string is not callable.

PR #198, opened 2019-05-08
Fix unit tests

dagwieers:
This PR includes:
- Add unit tests to pylint
- Fix unit tests (even if they don't work as-is yet)
- Run streamserviceests.py as part of unit tests (not by Travis)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ install:
 script:
 - tox
 - tox -e flake8
-- pylint *.py
-- pylint resources/lib/*/*.py
+- pylint *.py resources/lib/ test/
 - python test/vrtplayertests.py
+- python test/apihelpertests.py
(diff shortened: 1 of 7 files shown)


dagwieers:
@mediaminister @pietje666 Approved ?

mediaminister:
No! I get an error: `ImportError: No module named resources.lib.kodiwrappers`

dagwieers:
Ok, I fixed that.

pietje666:
Code wise it looks good :)

mediaminister:
I still get the same error. `ImportError: No module named resources.lib.kodiwrappers`

dagwieers:
@mediaminister Yup, this will require the same modifications as #187, so I think I am going to split off the unit test fixes, and drop this in favor of #187, which already has the required modifications (but not using the same path).

PR #199, opened 2019-05-08
Add new logging infrastructure

dagwieers:
This PR includes:
- A predefined set of log_levels: Quiet, Info, Verbose, Debug
- Default log_level is 'Info'
- Log all HTTP requests as Verbose
- Log all invocations as Verbose

This is only the first step of a set of troubleshooting improvements.

### DIFF
--- a/addon.py
+++ b/addon.py
@@ -28,6 +28,8 @@ def router(params_string):
 
     kodi_wrapper = kodiwrapper.KodiWrapper(_ADDON_HANDLE, _ADDON_URL, addon)
 
+    kodi_wrapper.log_notice('Path: ' + _ADDON_URL, 'Verbose')
+
     if action == actions.CLEAR_COOKIES:
         from resources.lib.vrtplayer import tokenresolver
         token_resolver = tokenresolver.TokenResolver(kodi_wrapper)
--- a/resources/lib/kodiwrappers/kodiwrapper.py
+++ b/resources/lib/kodiwrappers/kodiwrapper.py
@@ -11,18 +11,25 @@ try:
 except ImportError:
     from urllib import urlencode
 
-sort_methods = {
-    # 'date': xbmcplugin.SORT_METHOD_DATE,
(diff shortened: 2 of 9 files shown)


pietje666:
I do not really see how it is working, could you explain a bit more please

dagwieers:
It's not completely where I want this to go, but you can select Quiet|Info|Verbose|Debug in the settings, and this influences what is being logged. When using 'Info' you get the same logging as currently.

When set to 'Quiet', we don't log anything. When set to 'Verbose' you log every HTTP request and plugin invocation. When set to 'Debug' I plan to add a lot more things to be happening, potentially OSD information or notifications, a separate file with all played streams, a dump of the latest stream information, etc.

The idea is that with 'Debug' a user understands how the addon works roughly and can troubleshoot issues herself. That will be part of the unfinished 'Troubleshooting' section in the Wiki.

dagwieers:
Ah, and obviously we also need to take into account the Kodi log_level. So if you put the Kodi log_level to 'Debug', this affects the Addon logging as well.

pietje666:
I would just log using the default kodi loglevels, despite any addon setting. Also you look at it from your view as a developer but i think only a very small fraction of the users will use it

dagwieers:
@pietje666 Of course only a very small set of our users will be interested in this, but it's that group of users we want to understand what's going on and do great things with the stuff they learn. I don't want to have an addon that *we* have to keep on developing because people have no clue what's going on (or they need to be experienced/skilled in order to learn how things work). So I think this is critical for a good developer community that contribute amazing things.

The reason why only having the Kodi log_level, is that the stuff from the addon you are interested in will get lost with the stuff coming from Kodi potentially. So if people have a need to understand how the addon works, they do not have a need to find that in the inner workings of Kodi, that's a whole different ball game.

dagwieers:
Let's see where this goes. We can change it, or improve it before we release this to the public.

PR #202, opened 2019-05-08
implemented max bandwidth for hls streams

mediaminister:
Fix for https://github.com/pietje666/plugin.video.vrt.nu/issues/172


dagwieers:
Seems to work fine. Well done !

PR #203, opened 2019-05-08
Add 720p support without DRM

dagwieers:
I discovered that as soon as a query-string is added to a (live) stream_url, a 720p stream is part of the M3U8 file. This could be any string.

So this opens up the possibility to have 720p without the need for DRM.

### DIFF
--- a/resources/lib/vrtplayer/streamservice.py
+++ b/resources/lib/vrtplayer/streamservice.py
@@ -265,8 +265,8 @@ class StreamService:
         hls_audio_id = None
         hls_subtitle_id = None
         hls_base_url = master_hls_url.split('.m3u8')[0]
-        self._kodi_wrapper.log_notice('URL get: ' + master_hls_url, 'Verbose')
-        hls_playlist = urlopen(master_hls_url).read()
+        self._kodi_wrapper.log_notice('URL get: ' + master_hls_url + '?hd', 'Verbose')
+        hls_playlist = urlopen(master_hls_url + '?hd').read()
         max_bandwidth = self._kodi_wrapper.get_max_bandwidth()
         stream_bandwidth = None
 


dagwieers:
@mediaminister Can you test this ? If this is correct we may have to rethink how to document this in
https://github.com/pietje666/plugin.video.vrt.nu/wiki/HD-quality

I assume that this may fail to work for some content, not sure what to recommend for the future.

michaelarnauts:
@dagwieers I'm using a ?hd suffix for more then 6 months now with my stream in tvheadend. Works perfectly, no need for a login, no drm.

mediaminister:
You don't need to login to use the Live TV menu. That's a feature VRT supports, just like on https://www.vrt.be/vrtnu/livestream/

But adding a parameter like `?hd` is a hack. I think it's smarter not to document hacks. 
Hacks will stop working sooner or later.

If we want a working add-on in the long term we have to implement the features VRT supports. We have to implement DRM, login, etc...

dagwieers:
I don't think it is a hack, it's been added on purpose. The upside of not needing DRM for the live streams is that it will work better on less powerful hardware (i.e. Raspberry Pi). Not offering this is a problem IMO.

mediaminister:
It's a hack, test this: https://live-cf-vrt.akamaized.net/groupc/live/8edf3bdf-7db3-41c3-a318-72cb7f82de66/live_aes.isml/live_aes-audio_track_0_dut=128000-video=3000000-811196669.ts?itsahack

dagwieers:
Not sure what I need to test. If you mean that it works with any string, sure, that's what I said from the very start.

> I discovered that as soon as *a query-string* is added to a (live) stream_url, a 720p stream is part of the M3U8 file. This could be any string.

mediaminister:
Think again, VRT is not using DRM on 720p content because they like it. It's because if they want to show a Hollywood movie like "The King's Speech" next Saturday on Eén, they are contractually bound to use DRM if they distribute this movie over the internet.

dagwieers:
> Think again, VRT is not using DRM on 720p HD content because they like it.

@mediaminister Possibly I don't know. They aren't using 720p HD streams for the VOD content (both the content that they own, and the content that they assumingly need DRM for, which they could). So it's not as clear-cut. Your assumption is that this is an oversight at VRT's end? Which is something I don't believe. I have a high regard of VRT's technical competences. IMO This has been put in place on purpose. For what purpose, I can't say, but I don't think this is our concern if it is available to the public *and* being used widely already.

mediaminister:
I'm not convinced it's a feature because when you watch a livestream on https://www.vrt.be/vrtnu/livestream/ with an (old) browser that doesn't support DRM/Encrypted Media Extensions, the maximum quality you get is 540p.
I'm convinced VRT serves non-HD livestreams to devices that don't support DRM because they decided for certain reasons to do so.

dagwieers:
Time will tell, if it stops being provided, it stops working and our users go back to 540p.
I will leave the documentation as-is, unless you have any suggestions.

Can you already edit the wiki ? I don't have the rights to change the Wiki settings, or give you more rights to the Wiki though.

PR #206, opened 2019-05-10
Add search functionality

dagwieers:
In the future I like to extend this implementation:
- Show TV show listing, and if you click through, show the programs that matched the search query.
- Sort by score
- Store previous search queries


dagwieers:
In the future I like to extend this implementation by showing TV show listing, and if you click through, show the programs that matched the search query.

Doing it this way is more user-friendly and makes it possible to save such searches on a per-program basis.

Another feature would be to store previous searches, I need to look into how other plugins are doing this. Should not be hard to add.

dagwieers:
Merging for now, functional

PR #208, opened 2019-05-10
Add Ketnet Junior

### LINKED ISSUE #207: Add Ketnet Junior stream
For young children, Ketnet Junior can also be added as a live stream. They are using `https://content.uplynk.com/channel/e11a05356cc44198977436418ad71832.m3u8` for a few months now already, so it seems quite stable. They provide a nice 720p stream, no login required.

dagwieers:
This adds Ketnet Junior live stream.

This fixes #207

### DIFF
--- a/resources/lib/vrtplayer/__init__.py
+++ b/resources/lib/vrtplayer/__init__.py
@@ -54,6 +54,8 @@ CHANNELS = [
         name='ketnet-jr',
         label='Ketnet Junior',
         studio='Ketnet Junior',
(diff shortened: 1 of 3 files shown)


dagwieers:
@mediaminister I know you didn't like using StreamURLS directly, but otherwise it becomes even more complex. Too many abstraction-layers IMO.

mediaminister:
It's okay for now. But I fear VRT will phase out unprotected HLS streams in the near future.
When I find time, I will update streamservice so it will support Widevine DRM for Uplynk Video Streaming Service.

PR #211, opened 2019-05-10
Add setting to disabe menu caching

### LINKED ISSUE #210: Don't cache show listings
### Describe the bug ###

I have added shortcuts to some programs from the A-Z listing in Kodi. This makes it easy for the kids to start watching "Karrewiet", since they can just open it trough the favourites. They watch the latest episode every day, and go back to the main menu then. When they want to view the next episode the next day, they open "Karrewiet" again, but the listing doesn't refresh and the newest episode isn't visible. Going back and opening it again doesn't solve this.

A workaround is to add a different show, open that one, and then go back to karrewiet, but my kids say that's stupid :)

### To Reproduce ###
1. Add a show from the A-Z listing to favorites (context menu, add to favourites)
2. Go to the favourites and open the show.
3. Play something, or not, not important.
4. Go back to the main menu.
5. Time passes, a new episode has aired.
6. Go to favorites and open the show.
7. The old listing is displayed and no way to refresh is available.

### Expected behavior ###
When the show is opened, it should download the listing again.

### Additional context ###
- Operating system: OSMC
- Kodi version: 18
- Addon version: latest master
- Using a VPN: no
- Country you are using the addon from: Belgium

### Log (if available) ###
```paste below
N/A
```

dagwieers:
In some use-cases (e.g. when using favourites often) menu caching causes
new episodes to not appear when expected. In this case users can disable
this with an expected performance hit when navigating the menus.

This fixes #210 

This PR also changes the order of the Settings tabs to fix a long-standing issue related to the cursor being stuck in the email-field.

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -16,48 +16,52 @@ msgstr ""
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
 msgctxt "#30000"
-msgid "Credentials"
-msgstr "Credentials"
-
-msgctxt "#30001"
-msgid "Email address"
-msgstr "Email address"
-
-msgctxt "#30002"
-msgid "Password"
-msgstr "Password"
-
-msgctxt "#30010"
-msgid "Video"
-msgstr "Video"
-
-msgctxt "#30011"
-msgid "Subtitles"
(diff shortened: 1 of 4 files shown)


dagwieers:
cc @michaelarnauts

michaelarnauts:
Great!

PR #213, opened 2019-05-11
Add "My programs" menu

### LINKED ISSUE #181: 'Most Recent', but only of selected Favorites (Follow-support)
### Describe the feature ###

The 'Most Recent' list is great. But it still has a lot of items I don't care for.
So, the option of showing this list with only items that I have selected (Favorites) and I am actually following.

The selection maybe can be done via vrt.nu ("programma's die ik volg") or -maybe better- in Kodi (Favorites) itself.

This way, when the user comes home from work, he or she is presented with a sort of daily TODO-list: You see a list of all episodes of your favourite (selected) programs. Sometimes you skip a couple of days of watching TV or you get back from a longer holiday, and in the same way, you are automatically presented with an a bit longer list and you can start binch-watching to get back up to date.

Thus, the Most Recent list of the things I am following. (This is truly interactive and customised TV)

Note that there is no need to have folders per program or so. Just a flat chronological list of a mix of episodes TODO.

This is better than having to remember to set up your Telenet-box, for them to still mess up the timing so you still miss the last 10 minutes of your program anyway, like it is a VHS and 1991.

dagwieers:
This PR adds a way of tracking the programs you follow.
- Add a context menu to programs or episodes with either "Follow" or "Unfollow"
- Download and cache favorites from VRT NU
- Add an A-Z listing of followed programs
- Add a Recent listing of followed programs
- Add invisible setting 'usefavorites' to disable "My programs" (for unit tests)
- Disable brand filter in "My programs" recent listing (should be configurable really)

This fixes #181

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,7 @@ unit:
 	PYTHONPATH=$(pwd) python test/apihelpertests.py
 	PYTHONPATH=$(pwd) python test/tvguidetests.py
 	PYTHONPATH=$(pwd) python test/searchtests.py
+	PYTHONPATH=$(pwd) python test/favoritestests.py
 	@echo -e "$(white)=$(blue) Unit tests finished successfully.$(reset)"
 
 zip: test
--- a/addon.py
+++ b/addon.py
@@ -39,6 +39,21 @@ def router(params_string):
         _tvguide = tvguide.TVGuide(_kodiwrapper)
         _tvguide.show_tvguide(params)
         return
+    if action == actions.FOLLOW:
+        from resources.lib.vrtplayer import favorites
+        _favorites = favorites.Favorites(_kodiwrapper)
+        _favorites.follow(program=params.get('program'), path=params.get('path'))
+        return
+    if action == actions.UNFOLLOW:
+        from resources.lib.vrtplayer import favorites
+        _favorites = favorites.Favorites(_kodiwrapper)
+        _favorites.unfollow(program=params.get('program'), path=params.get('path'))
+        return
+    if action == actions.REFRESH_FAVORITES:
+        from resources.lib.vrtplayer import favorites
+        _favorites = favorites.Favorites(_kodiwrapper)
+        _favorites.update_favorites()
+        return
 
     from resources.lib.vrtplayer import vrtapihelper, vrtplayer
     _apihelper = vrtapihelper.VRTApiHelper(_kodiwrapper)
@@ -47,13 +62,15 @@ def router(params_string):
     if action == actions.PLAY:
         _vrtplayer.play(params)
     elif action == actions.LISTING_AZ_TVSHOWS:
-        _vrtplayer.show_tvshow_menu_items()
+        _vrtplayer.show_tvshow_menu_items(filtered=params.get('filtered'))
     elif action == actions.LISTING_CATEGORIES:
         _vrtplayer.show_category_menu_items()
     elif action == actions.LISTING_CATEGORY_TVSHOWS:
         _vrtplayer.show_tvshow_menu_items(category=params.get('category'))
     elif action == actions.LISTING_CHANNELS:
         _vrtplayer.show_channels_menu_items(channel=params.get('channel'))
+    elif action == actions.LISTING_FAVORITES:
+        _vrtplayer.show_favorites_menu_items()
     elif action == actions.LISTING_LIVE:
         _vrtplayer.show_livestream_items()
     elif action == actions.LISTING_EPISODES:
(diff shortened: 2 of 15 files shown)


dagwieers:
This is now full-featured, however lots of testing is needed before signing this off.

dagwieers:
@mediaminister @pietje666 I just noticed that due to the hardcoded `_API_KEY` my favorites are actually shared using the same profile. In effect all users would be managing the very same set of favorites.

dagwieers:
I am not sure what is going on here, it doesn't seem to be the _API_KEY. But I created a second user and somehow it is using the exact same favorites from VRT NU. (I cleared tokens and favorites cache)

dagwieers:
Nevermind, I think I fixed it :-)

mediaminister:
I know why this feature doesn't work. You need a different "new" X-VRT-Token.

The X-VRT-Token we use now to get the streams that require a login doesn't work with video-user-data.vrt.be
But this new X-VRT-Token has disadvantages in comparison with the old X-VRT-Token.
The old X-VRT-Token is one year valid and requesting one takes almost no time, just a couple http requests takes no more than 500 ms. With the cache of one year, it almost never slows down video playback.
The new X-VRT-Token is only one hour valid and it takes sometimes more than 1.5 seconds to request one. VRT uses many requests and callbacks between different pages for a simple login. A very bad design. With the new X-VRT-Token, video playback will definitely start slower, because every hour, you need a new X-VRT-Token.

For testing purposes only, I implemented the new "slow" X-VRT-Token to my test branch: https://github.com/mediaminister/plugin.video.vrt.nu/commit/6463b661cbb8bb27cf29ed556710cb6ac0a68279

But, I don't like the fact that the "My programs" feature will slow down video playback or menu loading. I noticed this feature requests an X-VRT-Token when starting the add-on. This blocks loading the main menu for seconds. I don't like that.

dagwieers:
@mediaminister Thanks for the investigative work.

However it should not slow down video playback or menu loading. As it only needs to be downloaded once every hour. And we can disable 'My programs' by default if that would be an issue.

BTW One option could be asynchronous downloads for cached objects. More caching (e.g. for A-Z listing and categories) is on my wishlist.

dagwieers:
I noticed your implementation replaced the original XVRTToken code, but I would separate them, and only use the short-lived token when it is needed.

dagwieers:
So I would like to merge this, sooner rather than later.
Once it is merged we can work on improving this code e.g. it doesn't impact users that don't use it.

Everything that's unrelated to "My programs" has been removed from this PR and merged in separate PRs.

mediaminister:
Still one big problem: "My programs" asks the user for credentials and requests an X-VRT-Token when starting the add-on. This blocks loading the main menu for seconds. This impacts all (new) users, especially those who don't use the "my programs" menu.
Don't forget that you don't need a VRT-login to watch the live streams. So it is totally unacceptable that we ask people for a VRT-login when this is not necessary.
 just use the add-on for livestreaming

dagwieers:
You are right, and it's easy to fix by only activating "My programs" when we have credentials _and_ an token. These things are fixable and I agree we shouldn't annoy users that only are interested in the live streams, or browsing the programs without ever playing one :-)

So let's fix it, rather than repeat the issues.

mediaminister:
That's a workaround. "My Programs" shouldn't make any requests to a server in the constructor. Especially when this is privacy sensitive information and the user has not initiated the request.
These requests should always be initiated by the user. For instance, when the menu is clicked or a setting is activated.

Just my point of view, I really don't like the fact that the add-on is downloading all my personal tv viewing data in a json file without me knowing or allowing it.

dagwieers:
> I really don't like the fact that the add-on is downloading all my personal tv viewing data in a json file without me knowing or allowing it.

1. It is not *all your personal TV viewing data*, it's only the favorites you selected.
2. The website is doing the exact same thing without opt-in or opt-out.
3. If you didn't select favorites, nothing is downloaded (I guess that is the opt-out).

Despite that, if you are concerned people are snooping the stuff your are viewing because of the JSON download, you are not trusting HTTPS. If that concern is warranted, you have a big problem using on-demand streaming, because that would be *actual TV viewing data* and is also using HTTPS.

If the concern is that these favorites are stored on your system, bad luck again, Kodi is storing even more *actual TV viewing data* already. And not just from VRT NU only.

So I don't agree with your assessment, your concerns are general issues related to on-demand streaming or maybe the security concerns with TLS. We could add a warning, but that would relate to the whole addon, not just the "My program" feature, but honestly, a warning would not be able to provide the right depth and balance.

PR #225, opened 2019-05-14
Make code more readable

dagwieers:
We have overly long constructions like:

    self._kodiwrapper.get_localized_string()

and this would now become:

    self._kodi.localize()

### DIFF
--- a/addon.py
+++ b/addon.py
@@ -26,38 +26,38 @@ def router(params_string):
     params = dict(parse_qsl(params_string))
     action = params.get('action')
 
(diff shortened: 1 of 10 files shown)


dagwieers:
This may be controversial, but it irritates me daily.

dagwieers:
For the help information in #231 I need to make some room in the language files.
So for this I plan to reorganize according to: https://github.com/pietje666/plugin.video.vrt.nu/wiki/Translations

So I need to merge this to avoid a very large conflict...

PR #231, opened 2019-05-15
Re-add option to disable favorites

dagwieers:
This PR includes:
- Re-added the settings option to enable/disable favorites
- Only enable My programs when settings enable it and credentials are detected
- Reorganize unit tests (move from vrtplayer to apihelper)

This is done over concerns from https://github.com/pietje666/plugin.video.vrt.nu/pull/213#issuecomment-492657267

Additionally, I think we need to add help to individual settings to give a little bit more insight into how settings affect the VRT NU addon. This also gives more maturity to our addon settings pages...

### DIFF
--- a/addon.py
+++ b/addon.py
@@ -39,25 +39,22 @@ def router(params_string):
         _tvguide = tvguide.TVGuide(_kodi)
         _tvguide.show_tvguide(params)
         return
+
+    from resources.lib.vrtplayer import favorites
+    _favorites = favorites.Favorites(_kodi)
     if action == actions.FOLLOW:
-        from resources.lib.vrtplayer import favorites
-        _favorites = favorites.Favorites(_kodi)
         _favorites.follow(program=params.get('program'), path=params.get('path'))
         return
     if action == actions.UNFOLLOW:
-        from resources.lib.vrtplayer import favorites
-        _favorites = favorites.Favorites(_kodi)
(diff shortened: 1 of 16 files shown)


dagwieers:
@mediaminister Does this help with your concerns, provided we add help information in the Settings.

dagwieers:
It appears that help messages don't work for addon settings, this is something for my list of Kodi shortcomings :-(

mediaminister:
Are you sure this works on a fresh Kodi installation? I can't use "my programs" on a fresh install.

My main concern is that a new function shouldn't slow down or break the basic functionality of the add-on (for new users):
These are my considerations:
- instantly showing the main menu
- livestreaming without login
- asking login, downloading user data only when needed and when the user initiates this request through a listitem (or setting).
- caching after the first user interaction, not before.

dagwieers:
@mediaminister If you do not have any credentials, the "My programs" will not appear. As soon as you enter your credentials (and have it enabled in the settings) that menu item should become visible. This is basically what you requested. People can now use the live streams and work as before without credentials.

All your considerations are met with this implementation, and it worked fine for me. I now also tested on Windows and discovered an issue with expirationDate when using Dutch Kodi locale (on an English Windows). I have a fix for this.

mediaminister:
Thanks for the explanation. It works now after providing credentials.
And many thanks for taking into account my concerns.

dagwieers:
I appreciate your feedback and help too ! :-)

PR #235, opened 2019-05-16
Fix spinner when not search term, or no results

dagwieers:
This was still an open issue with the current basic search functionality.


dagwieers:
I finally found the solution. Funny that both Netflix and Youtube plugin also have this problem...

Issue #238, opened 2019-05-16
Programs in TV guide no longer plays video


dagwieers:
### Describe the bug ###
I don't understand why, but it seems programs in the TV guide no longer can be played.

@mediaminister As you understand better how this mechanism works, can you investigate ?

mediaminister:
I'll look into it.

dagwieers:
I found the cause from #208: https://github.com/pietje666/plugin.video.vrt.nu/pull/208/files#diff-3068a65f7c424ae8246eb6d861867028R192

Not sure what the best way is to fix this.

mediaminister:
I have an updated streamservice ready that fixes this: https://github.com/pietje666/plugin.video.vrt.nu/pull/243

PR #243, opened 2019-05-16
improve streamservice

mediaminister:
- fixes https://github.com/pietje666/plugin.video.vrt.nu/issues/238
- simplify stream selection
- support for uplynk widevine drm
- improved hls substream selection

### DIFF
--- a/resources/lib/vrtplayer/vrtplayer.py
+++ b/resources/lib/vrtplayer/vrtplayer.py
@@ -4,7 +4,7 @@
 
 from __future__ import absolute_import, division, unicode_literals
 from resources.lib.helperobjects.helperobjects import TitleItem
-from resources.lib.vrtplayer import actions
+from resources.lib.vrtplayer import actions, streamservice, tokenresolver
 
 try:
     from urllib.request import build_opener, install_opener, ProxyHandler, urlopen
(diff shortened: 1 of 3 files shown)


resources/lib/vrtplayer/vrtplayer.py
```diff
@@ -4,7 +4,7 @@
 
 from __future__ import absolute_import, division, unicode_literals
 from resources.lib.helperobjects.helperobjects import TitleItem
-from resources.lib.vrtplayer import actions
+from resources.lib.vrtplayer import actions, streamservice, tokenresolver
```
dagwieers: Why import this at the top, when it may only be needed for playing a stream ?
(Whereas we import vrtplayer for almost everything else)

dagwieers:
This works really well, although it appears to not fix #238 in my tests.
Let's merge and do some more testing.

resources/lib/vrtplayer/vrtplayer.py
mediaminister: I thought this slowed down playback. Maybe I have to check this again.

dagwieers: It may, the very first time. The question is whether we want to import this when we start the addon for the first time (or every consecutive time after another addon has been opened), or whether we want to take this "hit" when we start the first playback.

I would expect it is only for the first playback, but if it is not, we should make these imports global so that it is. To be tested.

PR #248, opened 2019-05-17
Disable setIsFolder use (Kodi 18 specific)

dagwieers:
As reported, setIsFolder is new in Kodi 18, and we also support Kodi 17.

Thanks @mediaminister


michaelarnauts:
Can't this be wrapped in an if to check the Kodi version? Not sure about this line, but there might be cases where we want to do cool stuff in kodi 18 that isn't supported in Kodi 17.

dagwieers:
@michaelarnauts Sure it can, but I doubt it will make a difference.

We currently set this 'isFolder' state as part of adding the ListItem's to the Virtual Directory. So I am not sure what the plan is inside Kodi with this for the future (it would help if the documentation was a bit clear about new additions that add existing functionality), but we don't need it at the moment.

That is also why I added it as a FIXME. At some point in the future we want to revisit this.

dagwieers:
Going to merge it now so people interested in testing the master branch in Kodi 17 are not affected :-)

PR #254, opened 2019-05-18
Various small fixes

dagwieers:
This PR includes:
- Small fixes to descriptions so it looks better
- Rename Debug setting into Troubleshooting settings
- Fix EPG info in live TV menu when there is no transmission
- Add inpustream.adaptive as an optional requirement
- Reorganize addon settings

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -2,12 +2,11 @@
 <addon id="plugin.video.vrt.nu" name="VRT NU" version="1.9.0" provider-name="Martijn Moreel, dagwieers">
     <requires>
         <import addon="resource.images.studios.white" version="0.0.21"/>
-        <!-- import addon="inputstream.adaptive" version="2.0.19"/ -->
         <import addon="script.module.beautifulsoup4" version="4.5.3"/>
         <import addon="script.module.dateutil" version="2.6.0"/>
         <import addon="script.module.inputstreamhelper" version="0.3.3"/>
+        <import addon="inputstream.adaptive" version="2.0.20" optional="true"/>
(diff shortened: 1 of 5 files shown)


addon.xml
```diff
@@ -2,12 +2,11 @@
 <addon id="plugin.video.vrt.nu" name="VRT NU" version="1.9.0" provider-name="Martijn Moreel, dagwieers">
     <requires>
         <import addon="resource.images.studios.white" version="0.0.21"/>
-        <!-- import addon="inputstream.adaptive" version="2.0.19"/ -->
         <import addon="script.module.beautifulsoup4" version="4.5.3"/>
         <import addon="script.module.dateutil" version="2.6.0"/>
         <import addon="script.module.inputstreamhelper" version="0.3.3"/>
+        <import addon="inputstream.adaptive" version="2.0.20" optional="true"/>
```
dagwieers: @mediaminister I doubt this has an effect on Kodi 17, but with a little luck it might update inputstream adaptive if it is already installed. (Here's hoping...) Could you test this has no ill-effect on Kodi 17 ?

mediaminister:
I can test two VM's: Libreelec 8.2.5 and 8.0.2 which is Kodi 17.6 and Kodi 17.3

mediaminister:
I tested. The optional attribute has no effect at all. I doubt the optional attribute works for binary addons. I think we can rely on the inputstreamhelper dependency and remove the inputstream.adaptive dependency completely.

dagwieers:
Ok, I'll remove it in the next round of small fixes.

PR #255, opened 2019-05-18
troubleshouting setting to ignore inputstream.adaptive

mediaminister:
A troubleshouting setting  that ignores inputstream.adaptive add-on on Kodi.
This causes streamservice to always select HLS streams that are played directly in Kodi player.
This setting should always be enabled unless you are troubleshooting playback problems.

When disabled, this setting also ignores "Use DRM" setting, because inputstream.adaptive is needed for (Widevine) DRM


resources/settings.xml
```diff
@@ -22,5 +22,6 @@
         <setting label="30865" help="30866" type="action" id="refresh_favorites" action="RunPlugin(plugin://plugin.video.vrt.nu/?action=refreshfavorites)"/>
         <setting label="30867" help="30868" type="action" id="clear_tokens" action="RunPlugin(plugin://plugin.video.vrt.nu/?action=clearcookies)"/>
         <setting label="30869" help="30870" type="select" id="max_log_level" values="Quiet|Info|Verbose|Debug" default="Info"/>
+        <setting label="30871" help="30872" type="bool" id="useinputstreamadaptive" default="true"/>
```
dagwieers: We can also make this option visible only when inputstream.adaptive is installed.

I am also considering to add in this window a few text-fields that contain important debug information, e.g.
inputstream.adaptive version, inputstreamhelper version, Widevine information (if possible), etc.

ANd maybe add some section headers to group things together.

dagwieers:
This is working great! Thanks!

dagwieers:
Something we have to document better. Without **inputstream.adaptive** I need to set Maximum bandwidth to 4096kbps to get 720p, whereas with **inputstream.adaptive** I get away with setting it to 3072kbps. So maybe we should indicate ranges in the settings, or increase the values in the table to cover both.

mediaminister:
The bandwidth should not differ much. I used peak bandwidth setting in streamservice for HLS because this is a mandatory attribute in HLS manifest files.

I think inputstream uses average bandwidth for MPEG DASH.
You can consider this a bug.

I don't know what's the right choice, peak or average bandwidth.

dagwieers:
Ah, good to know. It would be best if they both behave similarly.
But more importantly, we should document it as such for clarity ;-)

@pietje666 Did you give @mediaminister Wiki rights ?
Or give me admin rights on the repo so I can do it ;-)

resources/lib/vrtplayer/streamservice.py
```diff
@@ -204,7 +204,7 @@ def get_stream(self, video, retry=False, api_data=None):
                 drm_stream = (vudrm_token or uplynk)
 
                 # Select streaming protocol
-                if not drm_stream and self._kodi.has_inputstream_adaptive_installed() or drm_stream and self._can_play_drm and self._kodi.get_setting('usedrm') == 'true':
+                if not drm_stream and self._kodi.has_inputstream_adaptive() or drm_stream and self._can_play_drm and self._kodi.get_setting('usedrm') == 'true':
```
dagwieers: If I remember it right **and** has precedence over **or** which means that all of the following conditions need to be met:
- not drm_stream
- has_inputstream_adaptive or drm_stream
- can_play_drm
- kodi.get_setting('usedrm') == 'true'

So if **drm_stream is False** the second item only relies on **has_inputstream_adaptive** to be True. So this condition can be simplified to:
```python
if not drm_stream and self._kodi.has_inputstream_adaptive() and self._can_play_drm and self._kodi.get_setting('usedrm') == 'true':
```

resources/settings.xml
dagwieers: Done.

resources/lib/vrtplayer/streamservice.py
dagwieers: I prefer to rewrite it like this:
```python
                if not drm_stream and self._kodi.has_inputstream_adaptive():
                    protocol = 'mpeg_dash'
                elif drm_stream and self._can_play_drm and self._kodi.get_setting('usedrm') == 'true':
                    protocol = 'mpeg_dash'
```
Still, I am not sure this is what is intended if we disable the use of inputstream.adaptive.

I am also not sure how MPEG_DASH or HLS related to inputstream.adaptive.

- Can we play both MPEG_DASH and HLS without inputstream.adaptive ?
- Can we play both MPEG_DASH and HLS with inputstream.adaptive ?

Would it make sense to also have an option like:
- Preferred streaming format: HLS / MPEG_DASH

Maybe we need a matrix in the Wiki to

mediaminister: 1) InputStream Adaptive is indispensable to play MPEG DASH (with or without Widevine DRM protection)
2) Widevine DRM-support (for InputStream Adaptive) is not available on Kodi 17 at all.
3) We can play HLS with Inputstream Adaptive but also with standard Kodi player
4) In Kodi 17 different older versions of InputStream Adaptive are available with less functionality.
5) There are 7(!) different Kodi 17 releases and it's unclear how well InputStream Adaptive works on all these versions. 

So, in streamservice I made choices to limit all possible streaming combinations:
- InputStream Adaptive installed: always play MPEG-DASH
- InputStream Adaptive is missing: always play HLS

This is a simple workflow and it guarantees video and audio will always play on every Kodi version and platform.

Making HLS or MPEG-DASH a user choice makes little sense. The `useinputstreamadaptive` setting basically has the same effect.

Splitting up the if statement for cosmetics is okay (both ` can_play_drm` and `has_inputstream_adaptive` functions listen to the `useinputstreamadaptive` setting.)

Documenting stream compatibility in the wiki won't hurt but is not a priority for me.

dagwieers: @mediaminister Thanks for the detailed description. It may be useful to detail this in the code, and maybe even in the settings when we disable InputStream Adaptive (e.g. this forces use of HLS) in the help (which isn't shown *yet*, needs fixing in Kodi).

dagwieers: @mediaminister I noticed you have been promoted to *Collaborator* so you can now assign things to me, or ask me to review something. Welcome :-)

PR #256, opened 2019-05-18
Add 'Install Widevine' settings option

dagwieers:
This PR enables the (re-)installation of Widevine for testing purposes.


dagwieers:
It seems our tests using a date (in this case the offline list) is sensitive to timezones. If CI is running in another timezone, today's date may be Belgium's yesterday. And obviously the offline list will be empty (as we are asking for a list of files that have already been removed).

This needs some re-thought, as it means the offline list may not work for people in other timezones correctly. We should always be returning CEST dates...

dagwieers:
So for offline matching I now use the **Europe/Brussels** timezone, rather than the local timezone.
And that seems to work fine. However we need to test this functionality on Windows and Android as well.

PR #258, opened 2019-05-20
Implement caching framework

### LINKED ISSUE #241: Add caching for A-Z listing and categories
### Describe the bug ###
Much like we now implemented the Favorites caching, we could do the same for other types of files. This implementation would not only offer speed improvements, it also offers redundancy.

Data illegible for caching could be:
- A-Z listing (TTL: 1 hour)
- Category listing (TTL: 1 week)
- TV guide information (e.g. yesterday, today, tomorrow) (TTL: 1 hour)
- Favorites (TTL: 1 hour)

The implementation would work like this:
1. Before downloading a new listing, check if a recent cache exists
   - Use the cache if recent (timestamp < TTL)
2. If cache is too old, attempt to download newer version
   a. If download failed, use cache anyway, update timestamp
   b. If download succeeded, hash content
      - If hash is different, save to disk
      - If hash is same, update timestamp

The troubleshooting settings would have an option to clear the caches.

dagwieers:
This PR includes:
- Caching and fallback framework for listings (e.g. A-Z, or categories)
- New troubleshooting option to invalidate caches

This fixes #241

### DIFF
--- a/test/xbmcaddon.py
+++ b/test/xbmcaddon.py
@@ -10,7 +10,7 @@ import sys
 
 # FIXME: Get information from addon.xml
 ADDON_INFO = {
-    'author': 'Martijn Moreels',
+    'author': 'Martijn Moreel',
     'changelog': '',
     'description': '',
     'disclaimer': '',
@@ -30,6 +30,7 @@ ADDON_INFO = {
 
 PO = polib.pofile('resources/language/resource.language.en_gb/strings.po')
 
+# FIXME: Maybe move this to test/userdata/settings.xml ?
 SETTINGS = {
     'username': 'qsdfdsq',
     'password': 'qsdfqsdfds',
@@ -43,7 +44,7 @@ SETTINGS = {
 
 # Read credentials from credentials.json
 if os.path.exists('test/credentials.json'):
-    SETTINGS.update(json.loads(open('test/credentials.json').read()))
+    SETTINGS.update(json.load(open('test/credentials.json')))
 else:
     print('Credentials not found in credentials.json', file=sys.stderr)
 
--- a/addon.py
+++ b/addon.py
@@ -29,6 +29,9 @@ def router(params_string):
     _kodi = kodiwrapper.KodiWrapper(_ADDON_HANDLE, _ADDON_URL, addon)
     _kodi.log_access(_ADDON_URL, params_string)
 
+    if action == actions.INVALIDATE_CACHES:
+        _kodi.invalidate_caches()
+        return
     if action == actions.CLEAR_COOKIES:
         from resources.lib.vrtplayer import tokenresolver
         _tokenresolver = tokenresolver.TokenResolver(_kodi)
@@ -52,7 +55,7 @@ def router(params_string):
         _favorites.unfollow(program=params.get('program'), path=params.get('path'))
         return
     if action == actions.REFRESH_FAVORITES:
-        _favorites.update_favorites()
+        _favorites.get_favorites(ttl=0)
         return
 
     from resources.lib.vrtplayer import vrtapihelper, vrtplayer
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -286,6 +286,10 @@ msgctxt "#30830"
 msgid "When enabled, menus are being cached so they work more quickly. Only in very specific use-cases this is problematic as new episodes may not appear when they are expected."
 msgstr ""
 
+msgctxt "#30831"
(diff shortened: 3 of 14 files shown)


dagwieers:
I would also like to add Recent/Offline caching as well. But the filtered content (My Recent/My Offline) needs special invalidation when things are followed/unfollowed.

Please test. I flagged it experimental for now.

mediaminister:
Does this speed up menu browsing? Didn't notice a difference with the master branch.

dagwieers:
It depends.

On Android I don't see a lot of improvements, except if I disable this caching some lookups are from time to time slower. (Likely network interruptions)

On Raspberry Pi the difference is bigger. What caching brings you is a more predictable load-time, but it definitely depends on the speed of the SD card, or whether it is still buffered in file system buffers. I noticed that even when things are cached, the JSON processing is the biggest culprit now.

And maybe in some cases getting it from the network may be faster than reading it from SD card. In this case you'd better replaced that SD card (I have had a LibreELEC that was quite slow due to the SD card wearing out).

Also, we have some use-cases where the same file is being loaded multiple times (e.g. tvguide if you first look what's on Eén, and then on Canvas it would be downloaded twice, or live tv, where it would get the EPG information 3 times when building the menu)

I will probably make a distinction between optional caching and mandatory caching. E.g. for categories if you would disable caching it will also not fall back to the categories cache, so for categories and the tvguide I would probably keep a separate option (or enable it always).

dagwieers:
@mediaminister Beware that for testing you *always* have to quit the addon, start another addon, and then go back to the addon. This is needed to ensure the new code has been loaded, since we enable **uselanguageinvoker** that's not a certainty anylonger.

dagwieers:
Just to indicate that the menu speeds on Raspberry Pi is bound to the JSON/menu processing, enable caching and go to the Categories menu. Then open and re-open the Kinderen-category (slow, 54 items) and the Koken-category (instant, 2 items).

We do caching for categories and channel listings as well. I wasn't planning to, but doing that change was pretty easy anyway and the results are even more visible (as the network is a larger part of the total load-time).

That is the reason why you don't notice e.g. the A-Z listing being a big speed up (something I expected). The cheer number of menu items slows everything down on Raspberry Pi.

(Slow here is quite relative, it's still close to a second here !)

dagwieers:
Recent and Offline is not being cached now, which I intend to do too.

dagwieers:
BTW I also intend to add a setting to disable fanart/thumbnails, as this may also play a part in slowing down menu listings (indirectly). Plus it helps me understand disk-access and network-access better (now this activity makes things less transparent).

dagwieers:
For offline and recent caching the differences (on Raspberry Pi) are also quite visible.
I only need to add a means for invalidating the filtered (My favorites) lists whenever programs are followed/unfollowed. But that's not that difficult now.

test/xbmcaddon.py
```diff
@@ -0,0 +1,69 @@
...
+
+from __future__ import absolute_import, division, print_function, unicode_literals
+import json
+import os
+import polib
+import sys
+
+# FIXME: Get information from addon.xml
+ADDON_INFO = {
+    'author': 'Martijn Moreels',
```
pietje666: not sure if deliberatly written wrong :D

dagwieers: It was pretty late and I planned to rewrite this anyway ;-)
As we can get all this information directly from the addon.xml anyway.

Will fix.

dagwieers:
> Why don't you use a mocking framework instead of writing your own mocks?

Well, if you look at what we did before, it was a kludge. And when I tried to improve it using Mock things were a lot harder then just writing proper modules for it. Also, mocking a library wasn't that easy and my aim was to import kodiwrapper as-is.

The xbmc libraries are not feature-complete, which is fine. If we ever add functionality we either need to add a stub, or add an alternative implementation.

PS This also opens up the possibility to run addon.py directly to achieve a higher coverage.

dagwieers:
So currently this is disabled by default, but I think we ought to enable this. If we ship this disabled, all our existing users will have it disabled. o if we enable it by default later, they will keep on having it disabled. What do you guys think ?

The TTL is 7 days for categories.json (with a fallback forever). The TTL for other files is 1 hour. And only for favorites.json we also use it as fallback forever. (This in case the favorites system would stop working, it would continue working locally, although updates would not be pushed :-/)

dagwieers:
Feedback welcome. In the meantime I'll merge it.

mediaminister:
I agree that we enable this by default, but before a release this feature must be thoroughly tested.

PR #267, opened 2019-05-23
Fix recalculation when using today, tomorrow, yesterday

dagwieers:
This fixes an issues where today, tomorrow and yesterday would not bring you to the advertised date between 12AM and 6AM.


dagwieers:
The only reason why we do not convert today, tomorrow and yesterday to a fixed date straight away, is that we want people to be able to bookmark this without having a fixed date.

PR #274, opened 2019-05-25
fix channels menu

mediaminister:


resources/lib/vrtplayer.py
```diff
@@ -110,7 +110,7 @@ def show_channels_menu_items(self, channel=None):
             self._kodi.show_listing(tvshow_items, sort='label', content='tvshows')
         else:
             from resources.lib import CHANNELS
-            channel_items = self._apihelper.get_channel_items(action=actions.LISTING_CHANNELS, channels=[c.get('name') for c in CHANNELS])
+            channel_items = self._apihelper.get_channel_items(action=actions.LISTING_CHANNELS, channels=[c.get('name') for c in CHANNELS if self._kodi.get_setting(c.get('name')) == 'true'])
```
dagwieers: This was not the intention. The channel selection was purely for filtering "Most recent" and "Soon offline".

addon.py
```diff
@@ -72,6 +72,7 @@ def router(params_string):
         _favorites.get_favorites(ttl=60 * 60)
         _vrtplayer.show_tvshow_menu_items(category=params.get('category'))
     elif action == actions.LISTING_CHANNELS:
+        _favorites.get_favorites(ttl=60 * 60)
```
dagwieers: I should check, I wouldn't expect the Channels menu listing to require favorites.
But I wouldn't expect it for the Category-listing either ;-)

dagwieers: Ok, I was wrong. We get the (filtered) program-listing through that same action.
So we do need it. That could be designed differently though...

I am going to add unit tests for the menu's as well, so we can actually run through all menu's to test those as well !

dagwieers: I fixed it slightly differently as in essence the top menu doesn't require favorites, only program-listing and episode-listings need favorites in order to have a Follow/Unfollow context menu item.

PS There is a possibility to add context menus through addon.xml based on ListItem properties. This could prove to be more efficient then adding one for each individual ListItem. To be tested.

resources/lib/vrtplayer.py
mediaminister: It's okay for me to adapt this, it just seemed a logical thing to do.

addon.py
mediaminister: Yesterday I noticed adding this separately to all menus slows down navigating (with caching disabled). There must be a more efficient way to fix this.

resources/lib/vrtplayer.py
dagwieers: I didn't even consider this. But I am open to change my mind.

addon.py
dagwieers: I hope so, disabling My programs should disable everything IMO. So this could definitely be enhanced.

On the other hand, I doubt people are really going to disable this, as I find it very useful. But it can work better, so it should.

PR #277, opened 2019-05-29
added new screenshots

mediaminister:


dagwieers:
Thanks !

PS Why are Ketnet and MNM not visible in that last image ? It may give the impression we do not offer it.

mediaminister:
I don't understand Ketnet and MNM live channels are visible. Maybe your image viewer is cutting of the image.

dagwieers:
Ah, I viewed them in GitHub, didn't notice they were cut off :-)
Sorry!

dagwieers:
BTW I prefer the screenshots also without parent directory. But that's just a small thing :-)

mediaminister:
I prefer following the Kodi add-on directory structure recommendations: https://kodi.wiki/view/Add-on_structure#resources.2Fmedia.2F

dagwieers:
I meant the parent directory (**..**) in the directory listings. I actually recommend to turn them off for a better experience: https://github.com/pietje666/plugin.video.vrt.nu/wiki/Kodi-settings#show-parent-folder-items--filelistsshowparentdiritems

mediaminister:
Okay, I was not aware of that setting. It doesn't matter to me.

dagwieers:
I know it's just a minor thing, but I have a feature request in my list to allow for fanart on those parent directory entries too. (i.e. that you can set fanart for the complete addon, and that every entry without working fanart would fall back to it).

PR #280, opened 2019-05-29
quick fix: only get favorites when activated

mediaminister:
The `usefavorites` setting doesn't work properly. When the setting is disabled, it is still calling the `get_favorites` function that blocks the menus when the user has no credentials provided.

This is a quick fix that solves the menu blocking.

### DIFF
--- a/resources/lib/favorites.py
+++ b/resources/lib/favorites.py
@@ -22,7 +22,7 @@ class Favorites:
         self._proxies = _kodi.get_proxies()
         install_opener(build_opener(ProxyHandler(self._proxies)))
         # This is our internal representation
-        self._favorites = None
+        self._favorites = dict()
 
     def is_activated(self):
         ''' Is favorites activated in the menu and do we have credentials ? '''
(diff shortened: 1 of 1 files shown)


dagwieers:
Hmm, I would rather do:
```python
    if not self.activated():
        return
```
and keep everything indented identically.

michaelarnauts:
@dagwieers indeed. This construct has a name, "guard clause". 

More info: https://refactoring.guru/replace-nested-conditional-with-guard-clauses

dagwieers:
@michaelarnauts Oh whoaw, I always try to find the exact explanation, but quickly give up. Need to remember that it's in these PRs ;-)

mediaminister:
Thanks for the explanation, I changed it.

dagwieers:
@mediaminister Thanks for fixing this. I was planning to look at it after v1.10.0 (mostly because I believe the majority of users like this functionality) but it is a bug nonetheless.

PR #288, opened 2019-06-03
Add more docstrings

dagwieers:
This adds file doc-strings.

But we should add them to tokenresolver.py, streamservice.py and helperobjects.py.


mediaminister:
I added docstrings for tokenresolver and streamservice in https://github.com/pietje666/plugin.video.vrt.nu/pull/289

PR #289, opened 2019-06-03
improve tokenresolver

mediaminister:
Changes:
- move all X-VRT-Token functionality from streamservice to tokenresolver
- add new refresh function, speeds up X-VRT-Token refresh
- add docstrings to tokenresolver and streamservice
- reduce duplicate code
- other small changes

### DIFF
--- a/resources/lib/router.py
+++ b/resources/lib/router.py
@@ -7,7 +7,7 @@
 from __future__ import absolute_import, division, unicode_literals
 
 import xbmcaddon
-from resources.lib import actions, kodiwrapper
+from resources.lib import actions, kodiwrapper, tokenresolver
 
 try:  # Python 3
     from urllib.parse import parse_qsl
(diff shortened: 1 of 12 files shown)


resources/lib/router.py
```diff
@@ -27,16 +27,15 @@ def router(argv):
     action = params.get('action')
 
     _kodi = kodiwrapper.KodiWrapper(addon_handle, addon_url, addon)
+    _tokenresolver = tokenresolver.TokenResolver(_kodi)
```
dagwieers: So I wanted to avoid loading TokenResolver even in the cases where it wasn't needed.
But maybe that was overly aggressive and there's no real need to. I didn't profile it.

resources/lib/streamservice.py
```diff
@@ -187,13 +186,13 @@ def _fix_virtualsubclip(manifest_url, duration):
                 manifest_url += '-' + end_time.strftime('%Y-%m-%dT%H:%M:%S')
         return manifest_url
 
-    def get_stream(self, video, retry=False, api_data=None):
-        '''Main streamservice function'''
+    def get_stream(self, video, roaming=False, api_data=None):
```
dagwieers: I like this better !

Issue #293, opened 2019-06-04
API/Command to play most recent entry of list


TASSDevon:
I’d like it to play the latest News entry in the News category/directory list with a command, so I can instruct Home Assistant through its Kodi module to do this.

mediaminister:
I made a pull request that implements a function that plays the latest episode of a tv show.
You can call this function through the [Kodi JSON-RPC API](https://kodi.wiki/?title=JSON-RPC_API)

For instance with curl:
```
curl -X POST -H "content-type:application/json" http://localhost:8080/jsonrpc -d '{"jsonrpc": "2.0", "method": "Player.Open", "params":{"item": {"file" : "plugin://plugin.video.vrt.nu/?action=playlatestepisode&tvshow=het-journaal" }}, "id" : "1"}'
```
I'm not familiar with Home Assistant but this should work with `media_player.kodi_call_method`

TASSDevon:
Works, thanks!

Here's the HA setup:

<img width="601" alt="Screenshot 2019-06-08 at 15 06 02" src="https://user-images.githubusercontent.com/6917072/59147703-1dbb7100-89ff-11e9-9572-2c15bfb64b32.png">

dagwieers:
Thanks for reporting back !

And thanks to @mediaminister for implementing it straight away ! :-)

dagwieers:
I renamed the **Autostart** wiki section to **Automation** and added the information from @mediaminister to that wiki page. Improvements and feedback welcome.

dagwieers:
@mediaminister Should we add a VRTNWS Live TV channel playing the latest episode in a loop ? That would be quite like the VRTNWS radio channel which is a continuous loop as well, or it is actually more like the VRTNWS radio stream channel which always starts from the beginning, but always the latest episode.

I think we should add it.

mediaminister:
I don't see the added value of a "Het Journaal" loop, there are plenty of possibilities in the menus to watch the last episode of "Het Journaal".

But VRTNWS occasionally offers live events, breaking news, press conference reports or "Facebook Live" sessions on https://www.vrt.be/vrtnws/nl/
I think we should use a VRTNWS Live TV menu item in the first place for for these events. Maybe you can open a new issue for to further investigate and discuss this functionality.

michaelarnauts:
I agree, I don't see the added value of a looped VRT NWS playlist.

dagwieers:
[...]

PR #294, opened 2019-06-04
added function to play latest episode of a tv show

mediaminister:
Fixes: https://github.com/pietje666/plugin.video.vrt.nu/issues/293


dagwieers:
That was quick :-) Thanks !

Issue #307, opened 2019-06-14
Some HLS streams fail with "415 Unsupported Media Type"


dagwieers:
### Describe the bug ###
There is an outstanding issue at VRT NU where some episodes do not work when using HLS streams (i.e. when not using Widevine). The problem has been reported a few times to VRT and they have confirmed this general issue. In most cases they will fix the faulty stream by hand when reported, but the root cause is yet unknown and unfixed.

### To reproduce ###
1. Disable Widevine (and restart the add-on)
2. Go to program **24 hours in police custody**
3. Click on episode **S04E05**
4. See error **415 Unsupported Media Type**

### Work around ###
Enable Widevine in the settings so the VRT NU add-on selects te MPEG-DASH stream that is unaffected by this issue.

### Additional context ###
- Operating system: LibreELEC 9.0.2
- Kodi version: Leia 18.2
- Addon version: 1.10.0
- Using a VPN: no
- Country you are using the addon from: Belgium

### Log (if available) ###
This is an example log from an earlier identical issue.
```paste below
2019-06-14 02:24:56.218 T:1321550704  NOTICE: [plugin.video.vrt.nu] Got cached token '/storage/.kodi/userdata/addon_data/plugin.video.vrt.nu/ondemand_vrtPlayerToken.tkn'
2019-06-14 02:24:56.219 T:1321550704  NOTICE: [plugin.video.vrt.nu] URL get: https://media-services-public.vrt.be/vualto-video-aggregator-web/rest/external/v1/videos/pbs-pub-cd09cc3c-5178-49c1-9387-715cb9ad98ff$
vid-d0b7cfaa-fd96-448e-9923-281b10261bfa?vrtPlayerToken=b10@b04c02c93691e93774a02025a289e44ce86e4ef8fc3f9d567d427964d37a1c4c&client=vrtvideo
2019-06-14 02:24:56.466 T:1321550704  NOTICE: [plugin.video.vrt.nu] URL get: https://remix-cf-vrt.akamaized.net/remix/18f7b87b-2715-490c-9ac6-8f05fbf5c7e4/remix_aes.ism/.m3u8?hd
2019-06-14 02:24:56.655 T:1321550704   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                             - NOTE: IGNORING THIS CAN LEAD TO MEMORY LEAKS!
                                            Error Type: <class 'urllib2.HTTPError'>
[...]

PR #308, opened 2019-06-14
provide user message when VRT NU HLS stream fails with HTTP 415

mediaminister:
This fixes https://github.com/pietje666/plugin.video.vrt.nu/issues/307


dagwieers:
Very nice !

Issue #306, opened 2019-06-14
Empty "Streaming" section when inputstream.adaptive is missing


dagwieers:
### Describe the bug ###
When **inputstream.adaptive** is missing from the system (vanilla Kodi Leia installation) the **Troubleshooting** settings has an empty **Streaming** section. We should probably hint to install **inputstream.adaptive** using a text field stating: *Please install or enable the **inputstream.adaptive** addon in Kodi*

PR #309, opened 2019-06-14
fix empty streaming setting section when inputstream.adaptive is missing

mediaminister:
This fixes https://github.com/pietje666/plugin.video.vrt.nu/issues/306


dagwieers:
THis is exactly what I had in mind !

PR #311, opened 2019-06-14
provide user message when VRT NU HLS stream fails with HTTP 415 (#308)

mediaminister:


resources/lib/streamservice.py
```diff
@@ -273,7 +284,12 @@ def _select_hls_substreams(self, master_hls_url):
         hls_subtitle_id = None
         hls_base_url = master_hls_url.split('.m3u8')[0]
         self._kodi.log_notice('URL get: ' + unquote(master_hls_url), 'Verbose')
-        hls_playlist = urlopen(master_hls_url).read().decode('utf-8')
+        try:
+            hls_playlist = urlopen(master_hls_url).read().decode('utf-8')
+        except HTTPError as e:
+            if e.code == 415:
```
michaelarnauts: Shouldn't you raise the exception again when not 415 for other errors?

dagwieers: Indeed. Can you edit the file and make a PR ?

mediaminister: Thanks for reporting, fixed in https://github.com/pietje666/plugin.video.vrt.nu/pull/315

PR #314, opened 2019-06-16
Clean url router

mediaminister:
This implements the new router design as discussed in https://github.com/pietje666/plugin.video.vrt.nu/issues/103

### DIFF
--- a/resources/lib/__init__.py
+++ b/resources/lib/__init__.py
@@ -117,38 +117,32 @@ CHANNELS = [
 ]
 
 
-class actions:
-    ''' A list of add-on actions '''
-    DELETE_TOKENS = 'deletetokens'
-    FOLLOW = 'follow'
-    INSTALL_WIDEVINE = 'installwidevine'
-    INVALIDATE_CACHES = 'invalidatecaches'
-    LISTING_ALL_EPISODES = 'listingallepisodes'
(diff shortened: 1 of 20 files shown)


dagwieers:
Oh god, I broke it really good. Need to think about how to add `refresh` in this new scheme.

dagwieers:
Let me offer a fix for this the coming week.
I think we can use the existing scheme and merge it with the CACHES database as it was implemented. I just need to add `@plugin.router()` entries for most of what is in **vrtplayer.py** and add my own for the refresh-part. (Trying to combine `refresh` with the existing logic is way to invasive, so I am not going to try that  with the new scheme).

mediaminister:
> Oh god, I broke it really good. 
I think i fixed it. Don't overcomplicate things.

dagwieers:
Well, I don't like the fact that we have an interface now for removing arbitrary files.
So let's merge this and then see how we can improve things.

dagwieers:
And thanks for fixing this. The quick succession of changes really speeds up development.

mediaminister:
Thanks for merging, now we can move forward with the new interface.
The cache refresh could surely be improved. For instance a container refresh doesn't refresh the tvguide caching file.

dagwieers:
@mediaminister I am rewriting using the routing module now and it is even looking more fabulous :-)
Hope I can test this in an hour and share it here.

PR #319, opened 2019-06-18
Add urls using routing

dagwieers:
Was planning to do it later, but why not now?

### DIFF
--- a/addon.py
+++ b/addon.py
@@ -12,13 +12,13 @@ import routing
 from resources.lib import kodiwrapper
 
 plugin = routing.Plugin()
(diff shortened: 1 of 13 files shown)


dagwieers:
There is one ugly bit in this PR, which is that we need to expose addon's (plugin.route-decorated) functions through globals() to the rest of the plugin. I'd rather have a better way to do this, but it seems that all other plugins only have a single module and then you don't have this  problem.

Not sure if there is a better way to do this (importing addon and referencing the methods did not work), maybe a change to the **routing** module that would allow a lookup based on function name (as we are doing now) would be a useful improvement.

Issue #305, opened 2019-06-13
Add "Featured content" menu (using programTags)


dagwieers:
### Describe the feature ###
The VRT NU website often shows different tags (programTags) with series or programs to promote them.

It would be nice if we could have a separate **Featured content** (Uitgelichte programma's) menu item which would offer each of these *special* recommendations, like:

- **[Tijdsgerelateerd](https://search.vrt.be/search?size=150&facets%5BprogramTags.parentTitle%5D=Tijdsgerelateerd)**
  - [Exclusief online](https://search.vrt.be/search?size=150&facets%5BprogramTags.title%5D=exclusief-online)
  - [Laatste aflevering](https://search.vrt.be/search?size=150&facets%5BprogramTags.title%5D=laatste-aflevering)
  - [Laatste kans](https://search.vrt.be/search?size=150&facets%5BprogramTags.title%5D=laatste-kans)
  - [Nieuw](https://search.vrt.be/search?size=150&facets%5BprogramTags.title%5D=Nieuw)
  - [Volledige reeks](https://search.vrt.be/search?size=150&facets%5BprogramTags.title%5D=volledige-reeks)
  - [Volledige seizoen](https://search.vrt.be/search?size=150&facets%5BprogramTags.title%5D=volledig-seizoen)
- **[Inhoudsgerelateerd](https://search.vrt.be/search?size=150&facets%5BprogramTags.parentTitle%5D=Inhoudsgerelateerd)**
  - [Kortfilm](https://search.vrt.be/search?size=150&facets%5BprogramTags.title%5D=kortfilm)

PR #320, opened 2019-06-18
More cleanups to new routing interface

dagwieers:
This PR includes:
- Adding docstrings to all routes
- Don't instantiate single-purpose objects
- Don't import libraries that are usd infequently in vrtplayer.py
- Remove test() function from addon.py for testing routing

### DIFF
--- a/addon.py
+++ b/addon.py
@@ -4,8 +4,6 @@
 
 ''' This is the actual VRT NU video plugin entry point '''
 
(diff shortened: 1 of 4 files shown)


dagwieers:
I am going to merge this now (I haven't tested it in full myself as I cannot now), but I like to work on #305 and I need this merged to avoid conflicts.

PR #328, opened 2019-06-19
Assorted fixes

### LINKED ISSUE #326: Watched Live channels are marked as watched
### Describe the bug ###
If you use one of the live channels once, the entry in the menu is marked as having been watched. This is a bit strange for a live channel, so we should look into disabling this overlay for live channels.

dagwieers:
This PR includes:
- Simplify /cache/delete interface
- Delay import json
- Make is_playable=False default in TitleItem
- Add playcount=0 InfoLabel for live streams so they are not watched
- Renamed "My A-Z" to "My favourite programs"
- Move My movies and My documentaries to the end of "My programs" menu
- Ensure unit tests are not over 160 rows
- Renamed "video_dict" to "info_dict"

This fixes #326

### DIFF
--- a/addon.py
+++ b/addon.py
@@ -24,10 +24,7 @@ def main_menu():
 @plugin.route('/cache/delete/<cache_file>')
 def delete_cache(cache_file=None):
     ''' The API interface to delete caches '''
-    if cache_file:
-        kodi.refresh_caches(cache_file)
-    else:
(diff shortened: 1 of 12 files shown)


dagwieers:
This did not used to be a problem since we changed the icon to the studio logo, but this is not working for the Channel-listing where we added the Live TV item.

PR #330, opened 2019-06-20
Improve Widevine/DRM settings text

dagwieers:
To avoid any confusion to when this option is required.


mediaminister:
Looks good to me, this is better and now even indispensable because VRT has been failing for weeks to provide a working HLS stream for _Appel d'urgence_ and maybe other protected VOD content.

PR #333, opened 2019-06-20
Do not clear username if login failed

dagwieers:
So currently when you type in a username and password, you can visually
verify the username is correct, but the password is a bunch of
asterisks. So if the login failed we assume the username was correct
(and can be verified by the user) but since the password is not visible
and cannot be edited we reset only the password.

This is a bit more userfriendly. An option to show the password would be
even better so we don't need to reset anything (just mark as invalid).

### DIFF
--- a/resources/lib/helperobjects.py
+++ b/resources/lib/helperobjects.py
@@ -39,7 +39,8 @@ class Credentials:
 
     def reset(self):
         ''' Reset the credentials in the settings '''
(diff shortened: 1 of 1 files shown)


michaelarnauts:
@dagwieers it might be useful to add an item to check the login from the settings below the username and password fields. When you don't know your password, it's quite a hassle to try a few combinations.

dagwieers:
@michaelarnauts Good idea! Something for the TODO list.

PR #334, opened 2019-06-20
Mix tvshows and oneoff episodes

### LINKED ISSUE #329: Merge directory items with oneoff video's
### Describe the feature ###
So one thing still bothering me is that in every view we show folders for single-item programs (typically indicated as **oneoff**). A much better user-experience would exist if we would perform a second query listing the oneoff's (could be cached as well) and merging that with the programs listing.

An implementation like this would work out of the box for the Movies and Short-movies listing as well so that they would not need to be exceptions.

dagwieers:
This implements showing oneoff programs in the list of tvshows. This
adds the convenience of not requiring to dig one level deeper to get to
the video.

This PR includes:
- Mixing multi-episode programs and one-off episodes (movies and documentaries)
- Add a one-off cache that is used for all oneoff-queries (number of oneoff programs is about 30)
- Indicate which programs are being followed using a °-character)

This fixes #329

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -119,7 +119,7 @@ msgid "My documentaries"
 msgstr ""
 
 msgctxt "#30045"
-msgid "All [I]oneoff[/I] documentaries on VRT NU"
+msgid "All [I]one-off[/I] documentaries on VRT NU"
 msgstr ""
 
 msgctxt "#30046"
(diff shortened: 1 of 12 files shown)


dagwieers:
This currently lacks caching as the Refresh-implementation does not work for multiple cache-files.
I plan to reinstate my earlier implementation to support this, unless something better is coined.

dagwieers:
After a weekend of testing I am going to merge this as-is.

In the unexpected case that 2 or more episodes for the same program are marked 'one-off', each episode will appear in the listing.

PR #335, opened 2019-06-24
Prepare for v2.0.0 release

dagwieers:

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,9 @@ VRT NU is the video-on-demand platform of VRT, Flanders' public broadcasting ser
 
 This Kodi add-on can be used to watch all video content from VRT NU.
 
-- Track the programs you follow
-- List all videos alphabetically by program, by category or by channel
-- Watch live streams from Eén, Canvas, Ketnet and Ketnet Junior
+- Track the programs you like
+- List all videos alphabetically by program, by category, by channel or by feature
+- Watch live streams from Eén, Canvas, Ketnet, Ketnet Junior and Sporza
 - Discover recently added or soon offline content
 - Browse the online TV guides or search VRT NU
 
@@ -49,6 +49,19 @@ You can report issues at [our GitHub project](https://github.com/pietje666/plugi
 you can send a message to [our Facebook page](https://facebook.com/kodivrtnu/).
 
 ## Releases
+
+#### v2.0.0 (2019-06-30)
+- Changes to API calls, THIS BREAKS EXISTING FAVOURITES AND WATCHED ITEMS
+- Added "Featured content" menu, with e.g. "Complete season" or "Last chance" (@dagwieers)
+- Added "My movies" and "My documentaries" menus (@dagwieers)
+- Added "Refresh" to context menu to refresh without cache (@dagwieers)
+- New API call to play latest episode of a program, i.e. home automation (@mediaminister)
+- Added Live TV entry at the top of channel-menu, convenient for favourites (@dagwieers)
+- Mix one-off programs with multi-episode programs (@dagwieers)
+- Indicate followed programs using the °-symbol (@dagwieers)
+- Improvements to the add-on settings and error messages (@mediaminister)
+- Store previous search queries (@dagwieers)
+- Small performance/efficiency improvements (@mediaminister)
 
 #### v1.10.0 (2019-05-30)
 - Performance improvements by caching HTTP requests locally (@dagwieers)
--- a/addon.xml
+++ b/addon.xml
@@ -22,9 +22,9 @@
         <description lang="en_GB">
 This Kodi add-on can be used to watch all video content from VRT NU.
 
-- Track the programs you follow
-- List all videos alphabetically by program, by category or by channel
-- Watch live streams from Eén, Canvas, Ketnet and Ketnet Junior
+- Track the programs you like
+- List all videos alphabetically by program, by category, by channel or by feature
+- Watch live streams from Eén, Canvas, Ketnet, Ketnet Junior and Sporza
 - Discover recently added or soon offline content
 - Browse the online TV guides or search VRT NU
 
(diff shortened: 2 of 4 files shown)


dagwieers:
So we don't have to rush this. I have a provisional date set of 2019-06-30.

Not sure how we want to announce the fact that this release breaks existing favourites and watched content. Should we provide a pop-up on first run after update (i.e. look for an unset new option or how can we detect an upgrade ?)

mediaminister:
I'm testing this release in the coming days. A pop-up on first run would be nice if possible. I'll will create pull requests with fixes if necessary. I'll finalize my testing on Saturday.

dagwieers:
@mediaminister As I said, there is no need to rush this. I'd rather get this right, so we don't ever have to break things for users again. And that's the only concern I have with this release, everything else we can improve/fix later, except if it would introduce breakage.

dagwieers:
@pietje666 @michaelarnauts Are you interested in helping to test the upcoming v2.0.0 release? Or already in holiday-mood? ;-)

dagwieers:
@mediaminister I was thinking of adding a hidden "version" field in the settings. And upon start check if it is different than the current version, show a pop-up window if not, and add the current version to that field. It is the easiest way to verify if we have upgraded from v1 to v2. (especially now since v1 has it unset). Later we can use this field for other similar needs if needed.

Would this work for you?

mediaminister:
@dagwieers Seems okay to me.

michaelarnauts:
@dagwieers I will try to do some tests tonight. My kids used to be a daily user of the app, bit since Karrewiet doesn't air in the holidays, this will probably change :)

michaelarnauts:
@dagwieers looks fine. To bad there isn't a way to "migrate" the old favorites to the new urls.

The other thing that bothers me (but this is not new on 2.0) is that on the main menu, the icon of the selected listitem is also the background (I guess this is because the same (low resolution) icon is set as `fanart`, like here: https://github.com/pietje666/plugin.video.vrt.nu/blob/6a9579086086bed1dcaef1894013e8263b9354da/resources/lib/vrtplayer.py#L39

I would drop that, since the icon looks really blurry when stretched in the background.

dagwieers:
@michaelarnauts There is no allowed way to migrate existing favourites (or what has been watched). Theoretically this is possible, but an add-on is not allowed to modify Kodi databases directly, and there is no interface to do this. What could be possible is a backward-compatibility layer that translates the old interface on to the new interface so existing favourites keep on working, but that's not a good solution IMO.

Wrt. the fanart, yeah, we do that on purpose. We can review this maybe when we have better artwork for the add-on that we can use as background, but at this time I prefer something blurry over nothing at all.

dagwieers:
I think we are almost ready, pop-up implemented. Raise any concerns now ! :-)

mediaminister:
There is a problem with `urllib2 unquote` and `quote` and unicode strings.
Following and unfollowing programs with non-ascii characters gives an error on python2 Kodi.
For instance: **Terug naar Siberië**

dagwieers:
I have a few things to discuss as well after [documenting the new routing on the Wiki](https://github.com/pietje666/plugin.video.vrt.nu/wiki/Automation):
- We have the following two routes to play by id
  - `/play/id/<video_id>`
  - `/play/id/<publication_id>/<video_id>`
    - *Would this not be better as: `/play/id/<video_id>/<publication_id>` ?*
- The route to play the latest episode of a program is now called:
  - `/play/lastepisode/<program>`
    - *Would this not be better as: `/play/latest/<program>` ?*

mediaminister:
For `/play/id/<video_id>` and `/play/id/<publication_id>/<video_id>` I followed VRT's own stream API scheme:
For on demand streams you need a publication_id and a video_id, for live streams you only need a video_id.
**On demand**
```
https://media-services-public.vrt.be/vualto-video-aggregator-web/rest/external/v1/videos/<publication_id>%24<video_id>?vrtPlayerToken=<vrt_player_token>&client=<client>
```
**Live**
```
https://media-services-public.vrt.be/vualto-video-aggregator-web/rest/external/v1/channels/<video_id>
```
No problem for me if you change this, but following VRT's stream API seemed logic.

I'm totally okay with changing `play/lastepisode/<program>` to  `/play/latest/<program>`

mediaminister:
I made some pull requests, apart from your requested internal api changes I think we're ready to release.
I tested already on LibreELEC 8.02, 9.02, Windows Kodi 19 Python3, Windows Kodi 18.2 and Linux Kodi 18.3.
I did no tests on Android and ARM platforms, that's a blind spot for now. Maybe in a couple of days I can test more. But if you want to release earlier, go ahead.

dagwieers:
I have tested on ARM last night and all was fine. I will do some more testing of your latest changes tonight on ARM. And if all goes well we can release tomorrow evening (still on 30/6 ;-)).

mediaminister:
Error on Android. Investigating as we speak.

dagwieers:
I tested this and it works fine in my case (recent started Kodi, not VRT NU, so when I started the updated VRT NU addon it wasn't already running/active).

But you have to know that you always have to restart the add-on after upgrading. That means you have to exit the add-on, load another add-on and then start the VRT NU add-on again. That will guarantee that the add-on is reloaded correctly. This is because of the **reuselanguageinvoker** option of the add-on, which speeds up the add-on.

dagwieers:
My only problem was that on my Android smartphone the pop-up title did not fit the window-width. So we need to rephrase/shorten that title-window.

dagwieers:
I wonder if there is a way to force the add-on to quit/restart when updated and **reuselanguageinvoker** is enabled. That should be default behaviour in Kodi actually.

mediaminister:
Indeed, reloading a freshly upgraded add-on is not done right in Kodi.

My suggestions for a shorter title:
Successfully installed VRT NU add-on 2.0.0
Welcome to VRT NU add-on 2.0.0

PR #339, opened 2019-06-27
Only order 'unsorted' when tvshows + oneoff

dagwieers:
This PR includes:
- Only order as unsorted when tvshows + oneoff
- Optional support for coloured studio icons
- Remove redundant 'icon' from setArt
- Order by episode when showEpisodeNumber is set


mediaminister:
Looks good to me.

dagwieers:
My only concern here is to both add "Episode X" to each episode, while also sorting them by episode (which adds an episode number). I think it would be better to remove the "Episode X" prefix.

mediaminister:
Indeed, in that case the episode prefix shouldn't be added.

PR #342, opened 2019-06-28
Fix UTF8 issue with quote()

dagwieers:
I was convinced I fixed this yesterday, but I probably forgot to backport it to my tree.


dagwieers:
I would rather have `urlopen()` to adhere to the `charset=UTF-8` part in the `Content-Type` header, so we don't have to decode bits and pieces ourselves.

PR #343, opened 2019-06-28
improve user interface strings

mediaminister:
This includes:
- fix typos
- grammar and vocabulary improvements


dagwieers:
Mostly ok, although I prefer a more formal writing in text: "Do not" over "Don't" or 'Is not" over "Isn't". This is how we learned to write English at school and I still prefer it for formal writing (i.e. also for documentation).

mediaminister:
Okay, I replaced n't with not.

PR #346, opened 2019-06-29
implement Unicode Sandwich

mediaminister:
This includes:
- always use unicode strings inside the add-on
- only decode/encode  when necessary for input and output on the outer edges of the add-on
 - fixes problems with special characters, for instance follow/unfollow Terug naar Siberië

### DIFF
--- a/resources/lib/vrtapihelper.py
+++ b/resources/lib/vrtapihelper.py
@@ -9,11 +9,11 @@ from resources.lib import CHANNELS, CATEGORIES, metadatacreator, statichelper
 from resources.lib.helperobjects import TitleItem
 
 try:  # Python 3
-    from urllib.parse import urlencode, quote, unquote
+    from urllib.parse import urlencode, unquote
     from urllib.request import build_opener, install_opener, ProxyHandler, urlopen
 except ImportError:  # Python 2
     from urllib import urlencode
-    from urllib2 import build_opener, install_opener, ProxyHandler, urlopen, quote, unquote
+    from urllib2 import build_opener, install_opener, ProxyHandler, urlopen, unquote
 
 
 class VRTApiHelper:
@@ -127,7 +127,7 @@ class VRTApiHelper:
         else:
             thumbnail = 'DefaultAddonVideo.png'
         if self._favorites.is_activated():
-            program_title = quote(tvshow.get('title').encode('utf-8'), safe=':/'.encode('utf-8'))
+            program_title = tvshow.get('title')
             if self._favorites.is_favorite(program):
                 context_menu = [(self._kodi.localize(30412), 'RunPlugin(%s)' % self._kodi.url_for('unfollow', program=program, title=program_title))]
                 label += ' [COLOR yellow]°[/COLOR]'
@@ -357,7 +357,7 @@ class VRTApiHelper:
 
         label, sort, ascending = self._make_label(episode, titletype, options=display_options)
         if self._favorites.is_activated():
-            program_title = quote(episode.get('program').encode('utf-8'), safe=':/'.encode('utf-8'))
+            program_title = episode.get('program')
             if self._favorites.is_favorite(program):
                 context_menu = [(self._kodi.localize(30412), 'RunPlugin(%s)' % self._kodi.url_for('unfollow', program=program, title=program_title))]
                 label += ' [COLOR yellow]°[/COLOR]'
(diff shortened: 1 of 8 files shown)


resources/lib/vrtapihelper.py
```diff
@@ -127,7 +127,7 @@ def tvshow_to_listitem(self, tvshow, program, cache_file):
         else:
             thumbnail = 'DefaultAddonVideo.png'
         if self._favorites.is_activated():
-            program_title = quote(tvshow.get('title').encode('utf-8'), safe=':/'.encode('utf-8'))
+            program_title = tvshow.get('title')
```
dagwieers: We still have to quote forward slashes in titles, otherwise it messes up the routing.

test/xbmc.py
```diff
@@ -33,7 +33,10 @@
         GLOBAL_SETTINGS = json.load(f)
 except Exception as e:
     print("Error using 'test/userdata/addon_settings.json' : %s" % e, file=sys.stderr)
-    GLOBAL_SETTINGS = {}
+    GLOBAL_SETTINGS = {
```
dagwieers: We ship this with the codebase, so in what case does this fail?

mediaminister: If someone deletes `test/userdata` completely and manually adds credentials.json
The wiki claims providing credentials.json is one of the two solutions to test:
https://github.com/pietje666/plugin.video.vrt.nu/wiki/Test-plan

resources/lib/vrtapihelper.py
mediaminister: Okay, I see, I'll fix this.

dagwieers: I already did last night, and the easiest was to create my own quote_slash()/unquote_slash() function.

test/xbmc.py
dagwieers: But I'm ok with this, I just wasn't expecting people to remove files we are shipping in the repository. (It also requires us to update this piece in sync with *test/userdata/addon_settings.json*)

mediaminister: When testing yesterday I removed the test/userdata folder because I didn't expect "userdata" to be shipped in the codebase. Maybe renaming the test subfolder names or documenting the Test plan wiki will prevent such mistakes.

dagwieers: I named it *userdata/* because the content resembles it. I was planning to also move *addon_settings.json* into *settings.xml* so that this file could also be copied over verbatim.

And I would prefer to move the token-files to a subdirectory *userdata/tokens/* too. As it would allow to remove all tokens at once.

But the additional files (i.e. global_settings.json and credentials.json) could go elsewhere. I simply could not think of a very good name for both. *test/config/* ?

mediaminister: So there is "userdata" before the user runs the codebase for the first time? I don't understand the logic of that.
In my opinion userdata should be generated on first run, just like Kodi does and just like cache and tokens are generated on first run.
Before the first run, there is no user, there are only "standard settings" which are shipped in the resources folder.

dagwieers: This is just for the unit tests. I am not interested to make the xbmc python libraries an exact copy of how XBMC behaves. If you want to improve it, be my guest, but IMO it's not worth the trouble (and could even be the cause of testing issues or false positives if this generating-code has bugs).

So yes, *userdata/addon_settings.json* is prefilled, and yes, there's no front-end to modify that file and we expect developers to manually edit that file to change it, although IMO we shouldn't recommend changing it on disk, but as part of the testing.

mediaminister: If generated "userdata" was separated from the standard shipped "userdata" I wouldn't have mistakenly removed a wrong folder. I just wanted to start with a clean test environment and didn't know how to do this.

dagwieers: The idea is that you can copy&paste an existing userdata/ directory from Kodi over our test/userdata/ and it would work. So that you don't need credentials, but you can use existing tokens and caches for testing. So yes, it works like this by design.

So as I said, the global_settings.json and credentials.json could move out of userdata/, but everything else in there is there on purpose. And only the stuff that is checked into the repository is there from the start and expected to be present (in whatever form). If you delete it, you can restore it using `git checkout test/userdata/addon_settings.json`. We provide a proper warning when this happens IIRC, but that may not have been tested properly.

dagwieers: Maybe `make clean` could also wipe cache, tokens and restore settings from repository?
But that could be considered data-loss if you modified them.

PR #348, opened 2019-06-29
improved 2.0.0 popup message

mediaminister:


resources/language/resource.language.en_gb/strings.po
```diff
@@ -670,5 +670,5 @@ msgid "A new major version was installed"
 msgstr ""
 
 msgctxt "#30979"
-msgid "Because of important internal changes, existing favourites from VRT NU no longer work, and the add-on no longer knows what was watched.\n[B]Please recreate all your VRT NU favourites.[/B]"
+msgid "Because of important internal changes, existing Kodi favourites and watched history from VRT NU no longer work.\n[B]Please recreate all your VRT NU favourites.[/B]"
```
dagwieers: Not all Kodi favourites are affected, only the ones that are related to VRT NU. That's what I wanted to stress.
But if you don't think people would be confused, I don't mind making it this concise.

PR #352, opened 2019-06-30
Add an action to check your credentials from the add-on settings.

### LINKED ISSUE #351: Allow checking credentials from an action in the settings xml
When configuring the addon, but you are not sure what your credentials are. The current flow is as follows:
* Go to settings and fill in your username and password.
* Save
* Open addon, go to A-Z, select a program and try to play it.
* An error occurs, start again.

We can use the `action` type in the settings xml to run an action that checks the credentials, and show a message if it failed or was successful.

I'm willing to create a PR for this, as an attempt to get to know the codebase a bit :)

michaelarnauts:
Fixes #351 

This adds a function to `tokenresolver.py` to just check the tokens. I couldn't reuse an existing function since they look in cache first. The credentials tab in the settings has an action now to run this function that checks the login credentials. A message is displayed indicating success or failure, and the settings dialog is shown again.

Note that I had to add `option="close"` to the settings since Kodi only saves the settings when the dialog is closed.

I also added `*.pyo` to the `.gitignore` file. Kodi creates these compiled files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,6 +72,7 @@ msbuild.wrn
 
 # Test caches
 *.pyc
+*.pyo
 .pytest_cache/
 .tox/
 
--- a/addon.py
+++ b/addon.py
@@ -35,6 +35,17 @@ def delete_tokens():
     tokenresolver.TokenResolver(kodi).delete_tokens()
 
 
+@plugin.route('/tokens/check-credentials')
+def check_credentials():
+    ''' Check if the credentials are correct '''
+    from resources.lib import tokenresolver
+    if tokenresolver.TokenResolver(kodi).check_credentials():
+        kodi.show_ok_dialog(message=kodi.localize(30980))
+    else:
+        kodi.show_ok_dialog(message=kodi.localize(30952))
+    kodi.open_settings()
+
+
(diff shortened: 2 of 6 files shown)


dagwieers:
I would like to postpone this to v2.0.1 or v2.1.0 (whatever comes first).

mediaminister:
Should definitely postponed because https://github.com/pietje666/plugin.video.vrt.nu/issues/351 is maybe not fully valid.

michaelarnauts:
Sure, I wasn't aiming at the 2.0 release

dagwieers:
I tested it yesterday and it works fine.

I am even inclined to remove the automatic clearing of the password when authentication fails. It would also be nice if there was an option to show the password (even if temporarily).

dagwieers:
So will the next version be v2.0.1 or v2.1.0 ? :-)
We are running out of ideas to implement...

michaelarnauts:
@dagwieers, i would prefer if the password won't reset. I had a typo in my username and it was a bit strange that my password was cleared :)

dagwieers:
@michaelarnauts Exactly my point :-) Do you want to make a PR for this?

PR #363, opened 2019-07-05
Add notifications to actions from settings

### LINKED ISSUE #362: Some Troubleshooting actions do not provide feedback
### Describe the feature ###
Currently if you clear your VRT tokens, refresh favorites or invalidate local HTTP caches in the Settings/Troubleshooting section, there is no feedback to the user that you activated that entry.

I think it could be worthwhile to give a notification when we perform any of these actions. I would refrain from giving a pop-up (which is more disturbing), but a notification gives direct feedback that something has happened.

This feedback is also useful when changing credentials (the only other place where this is activated), but we may want to detect when the credentials were actually changed, rather than clear VRT tokens on every access to settings.

dagwieers:
This PR includes:
- A notification when refreshing favorites
- A notification when clearing VRT tokens
- A notification when invalidating HTTP caches
- A notification when verifying credentials

This fixes #362

### DIFF
--- a/addon.py
+++ b/addon.py
@@ -44,11 +44,7 @@ def delete_tokens():
 def check_credentials():
     ''' Check if the credentials are correct '''
     from resources.lib import tokenresolver
-    if tokenresolver.TokenResolver(kodi).check_credentials():
-        kodi.show_ok_dialog(message=kodi.localize(30980))
-    else:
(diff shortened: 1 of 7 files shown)


dagwieers:
@mediaminister @michaelarnauts I turned the popup message box into a notification, which is less intrusive. Please test and report back!

mediaminister:
The notification works at first sight.

PR #364, opened 2019-07-06
Add YouTube support to VRT NU plugin

dagwieers:
This PR includes:
- A YouTube entry in the channels menu


dagwieers:
I was not sure if this was a good idea, but after adding it and seeing the vast amount of concerts from Radio 1 on YouTube made me realize this is an undiscovered diamant. Perfect to listen to concerts while working.

You can still disable the YouTube entries (i.e. if you do not want your children to end up inadvertently on YouTube).

dagwieers:
With this feature added, I think the next release will become v2.1.0 ;-)

Issue #367, opened 2019-07-06
Provide an API to start a program based on the timestamp.


michaelarnauts:
This issue is related to #209, but for the VOD content.

I use tvheadend to pass livestreams to kodi with a guide, so I have an overview of the programs of the past week or so in a nice UI. I'm now looking for a way to integrate a "net gemist"-like feature in this guide.

There is an easy way to create an extension that can add a context menu to the guide, and then pass the selected listitem to an python file. From there, it should be possible to find out what program was selected, on what channel, and trough a mapping of some kind, figure out what addon can be used to play this program. This would be something for another addon.

For vrt nu, this would only need an api endpoint that contains the channel, and the start time of the program. The addon could then lookup it up, and play the requested program.

This is a bit like the `playlatestepisode` feature from #294.

mediaminister:
What's the data source of the tvheadend TV guide? This is essential info because start times may differ between one TV guide and another TV guide depending on the data source.

For instance, Humo and VRT NU TV guides use different start times:
HUMO: https://www.humo.be/tv-gids
VRT NU: https://www.vrt.be/vrtnu/tv-gids/

michaelarnauts:
Hmm, didn't think of that. It seems that there can be a few minutes of. I'm currently using webgrab to scrape from telenet. 

Maybe a better way is to provide a start and end datetime, and see what program fits best in this timeslot? Since the start and end-times are only off a few minutes between guides, this right program can be selected quite easily. We can also use the program name to add to the matching, but that could also be a bit off.

mediaminister:
I implemented a basic version of this idea: https://github.com/pietje666/plugin.video.vrt.nu/pull/368

But there is a weak spot in this implementation: you always have to provide an exact air date. For instance, if you want to play Terzake of last friday, you should provide `canvas` and `2019-07-05T20:00:00` as parameters. A time stamp that differs only a second or minute from the timestamp won't work.

Example API request with curl
```
curl -X POST -H "content-type:application/json" http://localhost:8080/jsonrpc -d '{"jsonrpc": "2.0", "method": "Player.Open", "params":{"item": {"file" : "plugin://plugin.video.vrt.nu/play/airdate/canvas/2019-07-05T20:00:00" }}, "id": "1"}'

```

If you can provide an XML of your tv-guide, we can compare what data parameters match with the VRT NU api data.
[...]

PR #368, opened 2019-07-07
add method to play a program given a channel and airdate

mediaminister:


michaelarnauts:
@mediaminister you made a valid point in the issue regarding the different start times in different guides. I've replied here https://github.com/pietje666/plugin.video.vrt.nu/issues/367#issuecomment-508987131 that we maybe also should include the end time to best match a show.

dagwieers:
Matching the mean between start and end with the timeframe of broadcasts seems the most safe?
In fact we could support both, if only start is provided vs if both are provided.

michaelarnauts:
I'm not sure if it is possible to get the end time, so in that case the solution of @mediaminister seems to be the best choice.

mediaminister:
We can indeed support both, end date can be optional and this doesn't have to break the current implementation.

mediaminister:
Support for end_date is added.

dagwieers:
@michaelarnauts Can you test ?

@mediaminister When this is ready, please remove the WIP from the subject :-)

mediaminister:
It's not yet completely ready  but it works. Other improvements will be added at a later time.

dagwieers:
I need to know if this will go into v2.1.0 or not. I don't mind to add new functionality that may have issues (or could be improved). Especially since it is niche functionality. But what do you want?

mediaminister:
You can add this already to 2.1.0, no problem.

PR #370, opened 2019-07-07
Prepare for v2.1.0 release

dagwieers:
Update README and addon.xml for upcoming release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ VRT NU is the video-on-demand platform of VRT, Flanders' public broadcasting ser
 This Kodi add-on can be used to watch all video content from VRT NU.
 
 - Track the programs you like
-- List all videos alphabetically by program, by category, by channel or by feature
+- List all videos alphabetically by program, category, channel or feature
 - Watch live streams from Eén, Canvas, Ketnet, Ketnet Junior and Sporza
 - Discover recently added or soon offline content
 - Browse the online TV guides or search VRT NU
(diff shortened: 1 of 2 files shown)


dagwieers:
I expected to release v2.1.0 around July 15th, but now that we have an important bugfix, maybe we can test and release sooner?

mediaminister:
Maybe, if I find some time this week to finalize the airdate interface.

dagwieers:
So I guess we are ready to test and release. Since I haven't received any feedback related to the roaming VRT token, there is no urgency to get this out IMO.

dagwieers:
@pietje666 @michaelarnauts @mediaminister @dimsdupont Please test, and let me know when you're done testing. I have been using this release for the past few days without any issues.

mediaminister:
I found a python3 bug in `_get_roaming_xvrttoken`. I guess it's too late for 2.1.0

dagwieers:
I retrofitted the fix into v2.1.0, and recreated the ZIP file. It was not announced and is not distributed in the repositories yet. So no worries.

Since I don't see any new features on the TODO list or in the pipeline, I guess the next one will become v2.1.1.

PS I would like to get rid of the warnings from repo-plugins which suggests keeping addon.py and service.py small. So I probably will move the routing-stuff to router.py (or plugin.py ?). If you have any ideas around this, let me know.

mediaminister:
Thanks, I doubt the current livestream issue in #366 has anything to do with the add-on codebase. I tested everything yesterday with a Danish VPN and it only failed on the python3 bug.

For the next version, I will improve roaming error messages for users travelling outside Europe. Now, they get a wrong "please validate phone number" message.

Router.py is okay.
Maybe you can integrate https://pypi.org/project/kodi-addon-checker/ in the tests
pyo files should also be cleaned up, changing the Makefile
`	find . \( -name '*.pyc' -o -name '*.pyo' \) -type f -delete`

michaelarnauts:
I noticed a cosmetic issue with the airdate function. See #380. Not sure if it's worth holding back the release.

PR #372, opened 2019-07-08
improve airdate function

mediaminister:
This includes:
- added support for `RunPlugin()`
- Live-to-VOD: if VRT doesn't provide a VOD stream, you still can rewatch programs from the last 24 hours from the archived livestream

### DIFF
--- a/resources/lib/kodiwrapper.py
+++ b/resources/lib/kodiwrapper.py
@@ -233,7 +233,14 @@ class KodiWrapper:
             play_item.setSubtitles([video.subtitle_url])
 
         self.log_notice('Play: %s' % unquote(video.stream_url), 'Info')
-        xbmcplugin.setResolvedUrl(self._handle, bool(video.stream_url), listitem=play_item)
+
+        # To support video playback directly from RunPlugin() we need to use xbmc.Player().play instead of
+        # setResolvedUrl that only works with PlayMedia() or with internal playable menu items
+        if self._handle != -1:
+            xbmcplugin.setResolvedUrl(self._handle, bool(video.stream_url), listitem=play_item)
+        else:
+            xbmc.Player().play(item=video.stream_url, listitem=play_item)
+
         while not xbmc.Player().isPlaying() and not xbmc.Monitor().abortRequested():
             xbmc.sleep(100)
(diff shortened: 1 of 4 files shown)


resources/lib/streamservice.py
```diff
@@ -216,6 +216,10 @@ def get_stream(self, video, roaming=False, api_data=None):
             # Get stream manifest url
             manifest_url = next(stream.get('url') for stream in stream_json.get('targetUrls') if stream.get('type') == protocol)
 
+            # External virtual subclip support (live2vod airdate feature)
+            if video.get('start_date') and video.get('end_date'):
+                manifest_url += '?t=' + video.get('start_date') + '-' + video.get('end_date')
```
dagwieers: Nice !

resources/lib/streamservice.py
```diff
@@ -216,6 +216,10 @@ def get_stream(self, video, roaming=False, api_data=None):
...
+
             # Fix virtual subclip
             duration = timedelta(milliseconds=stream_json.get('duration'))
```
dagwieers: When using `get()` I always think about, what would happen if `duration` is not found?
Would milliseconds=None work, or do we need to provide a working default?
```
$ python -c 'from datetime import timedelta; print(timedelta(milliseconds=None))'
  File "<string>", line 1, in <module>
TypeError: unsupported type for timedelta milliseconds component: NoneType
```

dagwieers: Same for the other `get()` calls obviously.

resources/lib/kodiwrapper.py
```diff
@@ -229,7 +229,11 @@ def play(self, video):
             play_item.setSubtitles([video.subtitle_url])
 
         self.log_notice('Play: %s' % unquote(video.stream_url), 'Info')
-        xbmcplugin.setResolvedUrl(self._handle, bool(video.stream_url), listitem=play_item)
+        if self._handle != -1:
```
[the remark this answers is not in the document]
mediaminister: Done.

resources/lib/streamservice.py
mediaminister: Unfortunately, the livestream is only archived for 24 hours. I edited the comment accordingly.

PR #373, opened 2019-07-08
Add a disclaimer when the plugin is installed/updated

dagwieers:
This PR includes:
- A disclaimer for the end-user, so he is aware of the terms of use
- Implement a xbmcgui dialog textviewer
- Add the disclaimer to an About section in the settings

### DIFF
--- a/addon.py
+++ b/addon.py
@@ -231,6 +231,13 @@ def play_by_air_date(channel, start_date, end_date=None):
     vrtplayer.VRTPlayer(kodi).play_episode_by_air_date(channel, start_date, end_date)
 
 
+@plugin.route('/show/disclaimer')
+def show_disclaimer():
+    ''' Show the disclaimer '''
+    from resources.lib import vrtplayer
+    vrtplayer.VRTPlayer(kodi).show_disclaimer()
+
+
 if __name__ == '__main__':
     kodi.log_access(to_unicode(sys.argv[0]))
     plugin.run(sys.argv)
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -607,6 +607,14 @@ msgctxt "#30921"
 msgid "Log level"
 msgstr ""
 
+msgctxt "#30930"
+msgid "About"
+msgstr ""
+
+msgctxt "#30931"
+msgid "Show disclaimer..."
+msgstr ""
+
 
 ### MESSAGES
 msgctxt "#30951"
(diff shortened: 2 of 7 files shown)


dagwieers:
There are a few things to this:
- I want to ensure the end-user is aware of VRT's terms of use, and since the plugin could have been installed by a technical person, it's not sufficient to only show it on first use. That is why I opted to do it on every update.
- I prefered to use an ok-dialog, however this only show a very small windows with 4 lines and there is no way to influence the dimensions of the window
- Since we now basically have a textviewer, we could include the terms of use as well, however there is no text-version of the terms of use (AFAICT) so it means we need to scrape/format it ourselves. So maybe this is something we can do automated as part of making a new release, and ship the terms of use in the ZIP?

michaelarnauts:
My vote would be to only display on initial installation. Every update seems a bit much IMHO.

Non-technical people don't care if it's supported or not.

dagwieers:
A disclaimer is not about support.

mediaminister:
All VRT NU users received an e-mail message with all legal information after creating their VRT NU account. This is not our responsibility, we just provide the add-on "as is".

dagwieers:
I agree, but the disclaimer is specifically for the use of the VRT NU Kodi add-on.

At the time of creating the credentials, the user may not be thinking of the terms of use in light of using the VRT services with Kodi. So I am less concerned about the terms of use in itself, but instead how it affects the user when using our add-on.

If there is a concern that we are nagging the end-user, we could e.g. only do it for major and minor releases (no bugfix releases), or only do it once a month, every 3 months or 6 months (or progressively longer), or when the terms of use change (it means we have to scrape those etc.).

I am open to discuss every concern or proposal, but I think we need to make the end-user aware actively (rather than the current disclaimer in the README.md/addon.xml). And I think that's only possible if we do it repeatedly (i.e. not have a switch to disable it, as that can be automated to not appear).

mediaminister:
I don't understand what you're trying to fix here. The add-on shows already a short disclaimer in the add-on information window.

vrtnuaddon
https://kodi.tv/addon/plugins-video-add-ons/vrt-nu-0

I don't think it a good idea to abuse a TV user interface like Kodi to show long legal terms.
We can also refer to an online (wiki) page for a longer disclaimer.

If a user don't want to read long legal terms, that's not our responsibility: nemo censetur ignorare legem

dagwieers:
It would help if we could communicate in private to discuss concerns, but you don't answer your email. Not sure if you are even reading the email-address that is advertised on GitHub.

PR #378, opened 2019-07-10
fix python3 bug when roaming

mediaminister:
This fixes roaming on python3 Kodi


dagwieers:
So v2.1.0 was already cut, but the release has not been pushed to the Kodi repositories.
I decided to move this small bugfix into the released v2.1.0.

PR #384, opened 2019-07-15
Integrate pyformat with localize()

dagwieers:
By integrating string formattig with the localize() function call we can
make both the translations easier to comprehend, as well as the code
more transparant/readable.

And by using SafeDict() it is also more reliable if arguments do not add up.
When a key is not found the 'template' remains in the string.


dagwieers:
@mediaminister @michaelarnauts @pietje666 I wanted your feedback on this implementation, but since I require it for #385 I am going to merge it now. And we can improve/fix/change this as needed.

michaelarnauts:
I like this verry much. It's actually needed to translate to other languages since you don't know the order of the words.

mediaminister:
I am all in favor for more readable code.

PR #386, opened 2019-07-15
Add TV show name to context menu

dagwieers:
If you press Follow on an episode, you are not actually following the
episode, but the TV show. Make this more clear by adding the TV show
name to the context menu entry.

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -325,6 +325,10 @@ msgctxt "#30334"
 msgid "In 2 days"
 msgstr ""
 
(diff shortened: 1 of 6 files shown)


dagwieers:
Not a good idea as it becomes too long.
We may want to use **Follow program** or **Follow series** instead, not sure.

mediaminister:
It should be ** Follow program ** because someone can not only follow series but also one-offs.

dagwieers:
Well, for TV shows and one-offs we only use "Follow" anyway, since in this case it relates directly to the item that is selected. The ambiguity only exists for episodes.

PR #387, opened 2019-07-16
Various code structure changes

dagwieers:
This PR includes:
- Move addon.py to resources/lib/addon_plugin.py
- Move service.py to resources/lib/service_service.py
- Rename resources/lib/__init__.py to resources/lib/data.py
- Modify addon.xml to directly using resources/lib entry points
- Rename vrtapihelper.py to apihelper.py
- Fix a unicode issue in the TV guide
- Rename *show_&lt;foobar&gt;_menu_items* to *show_&lt;foobar&gt;_menu*
- Add kodi-addon-checker, but don't enable it by default
- Use branch-name in ZIP filename


mediaminister:
If addon.py only contains the router functionality, why don't you rename this file to router.py or navigation.py?

dagwieers:
That is what I did at first, but it also contains **plugin** and **kodi**. Using `router.plugin` or `router.kodi` looked rather strange.

mediaminister:
It's just a personal consideration, but using the general name **plugin** for the routing add-on is also rather strange.
We could import kodi at the add-on entry point and pass it to a  navigate function in our own navigation class.

dagwieers:
> We also could import kodi at the add-on entry point and pass it to a navigate function in our own navigation class.

That is what I prefer, but I haven't found an acceptable way to do that in combination with routing decorators.

dagwieers:
BTW I eventually moved it as addon.py also because for our tests it is just the same thing. It doesn't use default.py at all. I also contemplated using *resources/lib/addon.py* as entrypoint in *addon.xml*, but while it works, it may not get past the Kodi repo rules.

Also, we still need to do something similar for *service.py*, but honestly I don't see the point. Also here it makes more sense to point directly to *resources/lib/service.py* and also make the plugin using *resources/lib/* as a python path so everything becomes relative to the current directory.

mediaminister:
> but honestly I don't see the point.

This is the point: **the python files refered in addon.xml won't get compiled to bytecode**
This is explained here: https://github.com/xbmc/addon-check/issues/68

So, I think we can follow what [Twitch-on-Kodi](https://github.com/MrSprigster/Twitch-on-Kodi) did as a best practice

dagwieers:
> This is the point: the python files referred in addon.xml won't get compiled to bytecode

Well, we have `reuseLanguageInvoker` set. And our add-on is blazingly fast to start even with *all* the routing in that file which is not compiled to bytecode. So I don't think it matters that much. But I am not opposed to it either.

Thanks for the references, always nice to read :-)

I suggested doing what Twitch-on-Kodi is doing, but none of the add-ons in the official repositories are doing this. Same for moving the python path to resources/lib/, it works but nobody is doing it.

dagwieers:
So I moved *service.py* also to *resources/lib/service.py*.

mediaminister:
> We may want to ask the Kodi devs whether referring files in resources/lib/ from addon.xml would be acceptable

Youtube is doing this, so I think it's safe to do the same:
https://github.com/jdf76/plugin.video.youtube

dagwieers:
> Youtube is doing this, so I think it's safe to do the same

```xml
    <extension point="xbmc.python.pluginsource" library="resources/lib/default.py">
    <extension point="xbmc.service" library="resources/lib/startup.py" start="login"/>
    <extension point="xbmc.python.module" library="resources/lib/"/>
```

Oh, whoaw, this is exactly what I wanted. Does this also work for v17 ?
I only see one official plugin doing this currently in the krypton branch.

dagwieers:
@mediaminister This is the result.

dagwieers:
It appears that resources/lib/addon.py and resources/lib/service.py are also not stored as bytecode:
```
-rw-r--r--    1 root     root             0 Jul 16 23:46 __init__.py
-rw-r--r--    1 root     root          7388 Jul 16 23:46 addon.py
-rw-r--r--    1 root     root         38395 Jul 16 23:46 apihelper.py
-rw-r--r--    1 root     root         30403 Jul 16 23:46 apihelper.pyo
-rw-r--r--    1 root     root          5790 Jul 16 23:46 data.py
-rw-r--r--    1 root     root          4087 Jul 16 23:46 data.pyo
-rw-r--r--    1 root     root          6324 Jul 16 23:46 favorites.py
-rw-r--r--    1 root     root          7164 Jul 16 23:46 favorites.pyo
-rw-r--r--    1 root     root          2745 Jul 16 23:46 helperobjects.py
-rw-r--r--    1 root     root          4183 Jul 16 23:46 helperobjects.pyo
-rw-r--r--    1 root     root         26160 Jul 16 23:46 kodiwrapper.py
-rw-r--r--    1 root     root         27669 Jul 16 23:46 kodiwrapper.pyo
-rw-r--r--    1 root     root          3359 Jul 16 23:46 metadatacreator.py
-rw-r--r--    1 root     root          3141 Jul 16 23:46 metadatacreator.pyo
-rw-r--r--    1 root     root          4877 Jul 16 23:46 search.py
-rw-r--r--    1 root     root          4753 Jul 16 23:47 search.pyo
-rw-r--r--    1 root     root          1115 Jul 16 23:46 service.py
-rw-r--r--    1 root     root          4240 Jul 16 23:46 statichelper.py
-rw-r--r--    1 root     root          5500 Jul 16 23:46 statichelper.pyo
-rw-r--r--    1 root     root         18106 Jul 16 23:46 streamservice.py
-rw-r--r--    1 root     root         15170 Jul 16 23:46 streamservice.pyo
-rw-r--r--    1 root     root         15250 Jul 16 23:46 tokenresolver.py
-rw-r--r--    1 root     root         14284 Jul 16 23:46 tokenresolver.pyo
-rw-r--r--    1 root     root         12439 Jul 16 23:46 tvguide.py
-rw-r--r--    1 root     root         10757 Jul 16 23:47 tvguide.pyo
-rw-r--r--    1 root     root         13997 Jul 16 23:46 vrtplayer.py
-rw-r--r--    1 root     root         11725 Jul 16 23:46 vrtplayer.pyo
```
So it may not have brought us anything really.

mediaminister:
This statement is still true: **the python files refered in addon.xml won't get compiled to bytecode**
This means that `addon.py` and `service.py` never get compiled regardless of their location.

Just test with [kodi-addon-checker](https://pypi.org/project/kodi-addon-checker/)
```
WARN: Complex entry point. Check: resources/lib/addon.py | Counted lines: 182 | Lines allowed: 15
WARN: Complex entry point. Check: resources/lib/service.py | Counted lines: 23 | Lines allowed: 15
```
These two files should just import and call a method from another (compiled) library. Just like youtube did:

https://github.com/jdf76/plugin.video.youtube/blob/710822ded59fcc33ab1192189903bbb604b19189/resources/lib/startup.py#L11-L13

dagwieers:
Right, my hope was to have 2 useless files less in our tree.

Do you agree with the other changes in this PR?
I don't like the name of *data.py* but lack anything better.
An alternative is to store this information in JSON files on disk in resources/data/

mediaminister:
I have no other remarks.

mediaminister:
I got `unittest.loader.ModuleImportFailure` when running `make`.
Adding `export PYTHONPATH` to Makefile fixes this.

dagwieers:
> I got unittest.loader.ModuleImportFailure when running make.
That is unexpected, as PYTHONPATH is correctly set in Makefile and should therefore be available to all targets (albeit not exported). It does work for me. On what environment is this?

mediaminister:
Python 2.7.15+ on Linux Mint 19 (Tara)

dagwieers:
@mediaminister Vreemd, ik zou eerder denken dat het ligt aan `make`.
```
[dag@moria plugin.video.vrt.nu]$ rpm -q make
make-3.82-24.el7.x86_64
```

dagwieers:
From https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html

> To pass down, or export, a variable, make adds the variable and its value to the environment for running each line of the recipe. The sub-make, in turn, uses the environment to initialize its table of variable values. See Variables from the Environment.

Not sure how export would make a difference when `make` is putting PYTHONPATH in the environment of each recipe, unless pyunit is not passing it on to its own childs (in your case). Such a change seems quite fishy and error-prone.

dagwieers:
I added an export-statement in the Makefile, can you test if that works for you?
I don't think it should be required, but it doesn't harm either.

mediaminister:
I have make 4.1.
Everything works fine now with the export-statement.

dagwieers:
Is everything else in this mega PR fine?
Then I am going to merge this so I can start testing it with other branches.

mediaminister:
It's okay to merge.

PR #395, opened 2019-07-20
reorganize apihelper/metadata for airdate/latestepisode support

mediaminister:
This includes:
- Remove metadatacreator class
- Merge metadata functionality for different apis in new metadata class
- Reorganize and rename apihelper functions
- Metadata support for airdate and latest episode functions
- Fix bug in airdate function where night programs were not found
- Fix paging bug in search class


resources/lib/metadata.py
```diff
@@ -0,0 +1,584 @@
...
+
+        # VRT NU Search API
+        if api_data.get('type') == 'episode':
+            try:
+                season = int(api_data.get('seasonTitle'))
+            except ValueError:
+                try:
+                    season = int(api_data.get('seasonName'))
+                except ValueError:
+                    season = int()
```
dagwieers: This is a weird fallback value. Can you explain ?

dagwieers:
Looking good overall. I was not sure whether we wanted a different metadata class for search and suggest APIs, or do this, and I am still not sure what would be best. But this is definitely an improvement.

Do we want to return `int()`and `''` for unknown values, or instead return None?

resources/lib/metadata.py
mediaminister: I return an empty int because the `season` property  for the  `xbmcgui.ListItem` `infoLabels` should be an int. Other return values like `None` or` ''` are also possible though.

https://romanvm.github.io/Kodistubs/_autosummary/xbmcgui.html?highlight=setinfo#xbmcgui.ListItem.setInfo

mediaminister:
I just did what seemed logical to do. The previous implementation prevented me from easily adding the same metadata to multiple functions.

> Do we want to return int() and '' for unknown values, or instead return None?

Not sure if it matters, I should look into Kodi source code how this exactly differs.

mediaminister:
Can you merge this? I added the "year" functionality for movies and series.
I also checked Kodi source code, it doesn't matter what we return, Kodi's `ListItem::setInfo` uses C++ `strtol` to convert input to integers.
https://github.com/xbmc/xbmc/blob/770c691bfeea88d561b39d2e38348ac434db3adc/xbmc/interfaces/legacy/ListItem.cpp#L348

resources/lib/kodiwrapper.py
```diff
@@ -215,38 +215,42 @@ def show_listing(self, list_items, sort='unsorted', ascending=True, content=None
...
+            play_item.setProperty('subtitle', video.title)
+            play_item.setArt(video.art_dict)
+            play_item.setInfo(
+                type='video',
+                infoLabels=video.info_dict
+            )
         play_item.setProperty('inputstream.adaptive.max_bandwidth', str(self.get_max_bandwidth() * 1000))
         play_item.setProperty('network.bandwidth', str(self.get_max_bandwidth() * 1000))
-        if video.stream_url is not None and video.use_inputstream_adaptive:
+        if stream.stream_url is not None and stream.use_inputstream_adaptive:
```
dagwieers: Something to fix in the future: `stream.stream_url` &rarr; `stream.url`

resources/lib/kodiwrapper.py
```diff
@@ -215,38 +215,42 @@ def show_listing(self, list_items, sort='unsorted', ascending=True, content=None
...
-            self.log_notice('Subtitle URL: ' + unquote(video.subtitle_url), 'Verbose')
-            play_item.setSubtitles([video.subtitle_url])
+        if subtitles_visible and stream.subtitle_url is not None:
+            self.log_notice('Subtitle URL: ' + unquote(stream.subtitle_url), 'Verbose')
+            play_item.setSubtitles([stream.subtitle_url])
 
-        self.log_notice('Play: %s' % unquote(video.stream_url), 'Info')
+        self.log_notice('Play: %s' % unquote(stream.stream_url), 'Info')
 
         # To support video playback directly from RunPlugin() we need to use xbmc.Player().play instead of
```
dagwieers: This comment is no longer relevant.

dagwieers:
I see some room for improvement, but it's easier to propose that in a subsequent PR.
This is a major improvement over what we used to have. Thanks!

PR #406, opened 2019-07-30
fix extension point order

mediaminister:
This fixes an issue where the add-on is not visible in Chorus2 Web Interface and Embruary skin widget selection window.


dagwieers:
That is a weird issue and something that should probably be reported to both upstream too. Thanks for fixing!

Issue #404, opened 2019-07-27
When credentials change, favorites are not updated


dagwieers:
### Describe the bug ###
I just noticed that when you change your credentials, your favorites (and recent favorites + offline favorites) are not being refreshed, so you see the favorites from the previous credentials. Easy to fix.

mediaminister:
I'm working on a [fix](https://github.com/mediaminister/plugin.video.vrt.nu/tree/credentials) for this and meanwhile I found out https://github.com/pietje666/plugin.video.vrt.nu/commit/3c736c530056d2fa9397c4a310f0a8ba761443dc does not work and will never work because settings are changed after confirming the changes with the OK button.

Can anyone confirm my findings?

dagwieers:
I was planning to fix this issue as well, thanks for taking it.

I can confirm that it doesn't work within the settings, you have to first press OK to save it, and then go back. Maybe we ought to detect when credentials have changed, and do a test only then (with a notification or a pop-up).

michaelarnauts:
Strange. It should work since the option="close" saves the settings before closing the dialog. This is how I've seen it work.

dagwieers:
@michaelarnauts We don't close the dialog anymore, we show a notification.I guess that broke it.

michaelarnauts:
Ah. I missed that change. You need to close it. That's why the original code reopened the settings again after the check.

mediaminister:
I removed `check_credentials` completely on my test branch and implemented the same behaviour via the OK-button: https://github.com/mediaminister/plugin.video.vrt.nu/tree/credentials

dagwieers:
@mediaminister That's looking good. A nice cleanup as well. Perfect for the upcoming v2.2.0 ;-)

mediaminister:
Fixed in https://github.com/pietje666/plugin.video.vrt.nu/pull/410

PR #410, opened 2019-08-10
improve credentials handling

mediaminister:
This pull request includes:
- Fixed https://github.com/pietje666/plugin.video.vrt.nu/issues/404
- Reorganised tokenresolver to support idiot-proof credentials handling
- Removed class Credentials to simplify codebase
- Removed check_credentials because this became superfluous
- Improved version check for future use
- Added support for notifications for future use
- Other small changes related to credentials

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -411,14 +411,6 @@ msgctxt "#30806"
 msgid "Provide your VRT NU account password."
 msgstr ""
 
-msgctxt "#30807"
-msgid "Check credentials"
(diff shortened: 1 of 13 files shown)


dagwieers:
Oh no, this change again removes the password on login failure. And that is not what we want as it could be the email-address that is incorrect. (I now have to retype my very long password blindly :-/)

dagwieers:
Everything else worked fine. I will add a small fix to disable clearing the password.

mediaminister:
It's indeed better to let the user decide which input should be changed after showing an error message.

PR #416, opened 2019-08-11
New artwork

mediaminister:
This pull request includes:
- new artwork
- function to delete old artwork from cache on first run of 2.2.0 version


dagwieers:
My only concern here was the move from a `.jpg` to a `.png` because the documentation specifically talks about it being **fanart.jpg**. But looking at what exists in the official repository:
```
[dag@moria xbmc-repo-plugins]$ grep '.jpg</fanart>' */addon.xml | wc -l
109
[dag@moria xbmc-repo-plugins]$ grep '.png</fanart>' */addon.xml | wc -l
4
```

So I think we're good :-)

mediaminister:
kodi-addon-checker didn't complain and saving vector art in jpg format is completely insane.

PR #419, opened 2019-08-12
Use new high-res fanart by default

dagwieers:
This PR includes:
- Use the new high-res fanart instead of low-res icons
- Exclude .pyo files in Build-Zip
- Small changes to /tvguide rating
- Remove (debug) prints from test_kodi localize() tests


dagwieers:
Only remark, we now still use the low-res colour studio icons for channel-listings fanart (background). Not sure if we want to keep this or not.

dagwieers:
Let's merge this. We can do the channel-listings later if that's decided.

Issue #407, opened 2019-08-07
Add a "Go to program" context menu entry


dagwieers:
### Describe the feature ###
Make it easier to go to all episodes of a program from the **Most recent** or **Soon offline** menus. Currently if you discovered a new full season added in **Most recent**, there's no easy way to start with the first episode. Having a context menu to go to the program listing would fix this.

dagwieers:
Implemented in #425

PR #425, opened 2019-08-13
added go to program context menu

mediaminister:
This pull request implements https://github.com/pietje666/plugin.video.vrt.nu/issues/407

### DIFF
--- a/resources/lib/addon.py
+++ b/resources/lib/addon.py
@@ -105,6 +105,7 @@ def favorites_refresh():
 @plugin.route('/programs')
 @plugin.route('/programs/<program>')
 @plugin.route('/programs/<program>/<season>')
+@plugin.route('/programs/<program>/<season>/')
 def programs(program=None, season=None):
     ''' The programs A-Z / seasons / episode listing '''
     from vrtplayer import VRTPlayer
--- a/resources/lib/metadata.py
+++ b/resources/lib/metadata.py
@@ -88,6 +88,16 @@ class Metadata:
                         'RunPlugin(%s)' % self._kodi.url_for('follow', program=program, title=program_title)
                     ))
         context_menu.append((self._kodi.localize(30413), 'RunPlugin(%s)' % self._kodi.url_for('delete_cache', cache_file=cache_file)))
+
+        # Go to program context menu
+        if cache_file and cache_file.startswith(('offline', 'recent')):
+            season = self.get_season(api_data)
+            if season:
+                plugin_url = self._kodi.url_for('programs', program=program, season=season)
+            else:
+                plugin_url = self._kodi.url_for('programs', program=program)
+            # NOTE: ActivateWindow doesn't handle urls without trailing slash right
+            context_menu.append((self._kodi.localize(30417), 'ActivateWindow(Videos,%s,return)' % plugin_url))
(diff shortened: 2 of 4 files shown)


resources/lib/addon.py
```diff
@@ -105,6 +105,7 @@ def favorites_refresh():
 @plugin.route('/programs')
 @plugin.route('/programs/<program>')
 @plugin.route('/programs/<program>/<season>')
+@plugin.route('/programs/<program>/<season>/')
```
dagwieers: This means any constructed URL within the add-on will have a trailing slash.
Is this what we want?

resources/lib/metadata.py
```diff
@@ -88,6 +88,16 @@ def get_context_menu(self, api_data, program, cache_file):
                         'RunPlugin(%s)' % self._kodi.url_for('follow', program=program, title=program_title)
                     ))
         context_menu.append((self._kodi.localize(30413), 'RunPlugin(%s)' % self._kodi.url_for('delete_cache', cache_file=cache_file)))
+
+        # Go to program context menu
+        if cache_file and cache_file.startswith(('offline', 'recent')):
```
dagwieers: That's a clever way of doing this.

When investigating this, I gave up after realising it meant I had to forward even more information to helper functions and didn't think it was worth all the effort/complexity.

resources/lib/addon.py
dagwieers: I think if we swap the position of both, this is no longer an issue.

mediaminister: I added this route because I found out Kodi built-in `ActivateWindow` doesn't handle urls without a trailing slash right. I didn't found another way to fix this.

dagwieers:
Let's merge anyway.

resources/lib/addon.py
dagwieers: Because now it creates routes with trailing slashes internally:
```
» Season 2019 → /programs/pano/2019/
» Season 2018 → /programs/pano/2018/
» Season 2017 → /programs/pano/2017/
» Season 2016 → /programs/pano/2016/
```

dagwieers:
It appears that going back doesn't go back to the previous window (i.e. the recent/offline listing).
That is unfortunate, but I don't see how to fix it :-/

resources/lib/addon.py
mediaminister: If I change the order I get `No route to path` for all plugins url that are handled by `ActivateWindow`.

mediaminister:
> It appears that going back doesn't go back to the previous window

This might be a bug with `ActivateWindow`:
https://forum.kodi.tv/showthread.php?tid=336929

dagwieers:
I opened it as a bug here: https://github.com/xbmc/xbmc/issues/16492
We may want to track this separately in our issues too.

resources/lib/addon.py
dagwieers: Looking into this deeper, the problem seems to only exist when the season consists of numbers (e.g. `2016`), but not if it is a string (e.g. `allseasons`).

dagwieers: I reported this `ActivateWindow()` issue upstream at:: https://github.com/xbmc/xbmc/issues/16516

PR #427, opened 2019-08-13
Fix features/categories ordering in en_gb

dagwieers:
This PR includes:
- Fix the ordering of categories and features in en_gb
- Put channel TV guide entry in Bold in channel listing

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -310,7 +310,7 @@ msgid "TV guide for channel [B]{label}[/B]"
 msgstr ""
 
 msgctxt "#30303"
-msgid "[B]{label}[/B] TV guide"
+msgid "{label} TV guide"
 msgstr ""
 
 msgctxt "#30330"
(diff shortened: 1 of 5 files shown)


resources/lib/addon.py
```diff
@@ -104,8 +104,8 @@ def favorites_refresh():
 
 @plugin.route('/programs')
 @plugin.route('/programs/<program>')
-@plugin.route('/programs/<program>/<season>')
 @plugin.route('/programs/<program>/<season>/')
+@plugin.route('/programs/<program>/<season>')
 def programs(program=None, season=None):
```
mediaminister: This breaks urls handled by `ActivateWindow`. I get `No route to path`.

dagwieers: I'll remove it from this PR, but maybe we have to fix proper trailing slash handling in the routing library instead :-(

dagwieers: And we probably want integration tests for trailing slash handling as well.

PR #428, opened 2019-08-13
Bring together all caching settings

dagwieers:
This moves the menu and http caching into the Troubleshooting settings.


dagwieers:
I think overall it is better, but I don't like the order of the cache settings now :-)
You can never win...

mediaminister:
Looks good to me.

dagwieers:
My biggest issues is that actions in that menu aren't clear. If we go to outside menus we use a trailing `...`, but for actions it would be nice if it had some clear visual indication as well.

Issue #429, opened 2019-08-13
Clearing tokens disable My Programs (favorites)


dagwieers:
### Describe the bug ###
So I discovered that a user would lose his My Programs menu when he clears his tokens. And there's no way to get your My Programs back, unless you first invalidate you login (use a wrong login) and then change it back again in order to force a new login token.

### To Reproduce ###
1. Go to Settings » Troubleshooting
2. Click on Clear VRT tokens
3. Go back to main menu
4. See My programs is missing, and nothing will bring it back

### Expected behavior ###
My programs remains listed even when tokens have been cleared.

### Additional context ###
- Operating system: LibreELEC v9.0.2
- Kodi version: 18.3
- Addon version: 2.2.0
- Using a VPN: no
- Country you are using the addon from: BE

mediaminister:
You can easily get your "My Programs" back by starting an on demand stream.
I'll see how to fix this.

PR #430, opened 2019-08-13
fix my programs menu

mediaminister:
This fixes https://github.com/pietje666/plugin.video.vrt.nu/issues/429

### DIFF
--- a/resources/lib/favorites.py
+++ b/resources/lib/favorites.py
@@ -24,8 +24,7 @@ class Favorites:
 
     def is_activated(self):
         ''' Is favorites activated in the menu and do we have credentials ? '''
-        from tokenresolver import TokenResolver
-        return self._kodi.get_setting('usefavorites') == 'true' and self._kodi.credentials_filled_in() and TokenResolver(self._kodi).logged_in()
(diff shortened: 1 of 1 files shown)


resources/lib/favorites.py
```diff
@@ -24,8 +24,7 @@ def __init__(self, _kodi):
 
     def is_activated(self):
         ''' Is favorites activated in the menu and do we have credentials ? '''
-        from tokenresolver import TokenResolver
-        return self._kodi.get_setting('usefavorites') == 'true' and self._kodi.credentials_filled_in() and TokenResolver(self._kodi).logged_in()
+        return self._kodi.get_setting('usefavorites') == 'true' and self._kodi.credentials_filled_in()
```
dagwieers: Great, I didn't want to mess with that as I did not fully understand why this was there.

mediaminister: It was there to prevent the follow/unfollow function from failing if the user filled in wrong credentials and cancelled login.
But, this is a very rare situation and if it happens the user also gets a notification.
There is room to improve user experience in such rare cases, but this is not a priority.

PR #432, opened 2019-08-13
Move tokens to userdata/tokens/

dagwieers:
This PR includes:
- Move tokens to tokens/ subdirectory
- Clean up old tokens temporarily


dagwieers:
Not very elegant, but at some point in the future we can get rid of the old tokens.

mediaminister:
Tested and approved.

Issue #440, opened 2019-08-14
Livestream does not play because Widevine DRM fails


mediaminister:
### Describe the bug ###

The livestreams of Eén, Canvas, Ketnet and Sporza fail to play when "Use Widevine DRM (for protected content)" is enabled in the add-on settings. Playback tab

### To Reproduce ###
1. Make sure InputStream Adaptive add-on is installed in Kodi and "Use Widevine DRM (for protected content)" is enabled in the VRT NU add-on settings
2. Open VRT NU add-on and go to Live TV
3. Click Eén Live, Canvas Live, Ketnet Live or Sporza live
4. Kodi player tries to start the stream but halts immediately
5. The livestream doesn't play

### Expected behavior ###

When you click a livestream it should play just fine.

### Additional context ###

- Operating system: Linux Mint 19 Tara (based on Ubuntu 18.04.3 LTS (Bionic Beaver))
- Kodi version: 18.3
- Addon version: 2.2.0 
- Using a VPN: no
- Country you are using the addon from: Belgium

### Log (if available) ###
```paste below
2019-08-14 14:11:36.592 T:140514435241728   DEBUG: Curl::Debug - HEADER_IN: HTTP/1.1 403 FORBIDDEN
2019-08-14 14:11:36.592 T:140514435241728   DEBUG: Curl::Debug - HEADER_IN: Access-Control-Allow-Headers: Content-Type
2019-08-14 14:11:36.592 T:140514435241728   DEBUG: Curl::Debug - HEADER_IN: Access-Control-Allow-Methods: GET,POST,OPTIONS
2019-08-14 14:11:36.592 T:140514435241728   DEBUG: Curl::Debug - HEADER_IN: Access-Control-Allow-Origin: *
2019-08-14 14:11:36.592 T:140514435241728   DEBUG: Curl::Debug - HEADER_IN: Access-Control-Expose-Headers: Vualto-Transaction-Id
2019-08-14 14:11:36.592 T:140514435241728   DEBUG: Curl::Debug - HEADER_IN: Content-Type: application/json
2019-08-14 14:11:36.592 T:140514435241728   DEBUG: Curl::Debug - HEADER_IN: Server: TornadoServer/4.0.2
2019-08-14 14:11:36.592 T:140514435241728   DEBUG: Curl::Debug - HEADER_IN: Vualto-Transaction-Id: wv:21c18cf2-fd96-451f-a70c-6f2ad3530dbc
[...]

PR #442, opened 2019-08-15
Changes

mediaminister:
This pull request includes:
- require new inputstreamhelper 0.3.5 fixing https://github.com/pietje666/plugin.video.vrt.nu/issues/440
- new fix for webscraper after VRT NU changed html again

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -7,7 +7,7 @@
         <import addon="script.module.beautifulsoup4" version="4.5.3"/>
         <import addon="script.module.dateutil" version="2.6.0"/>
         <!-- import addon="inputstream.adaptive" version="2.0.20" optional="true"/ -->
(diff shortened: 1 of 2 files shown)


dagwieers:
Should we release v2.2.1 before v2.2.0 is out ?

mediaminister:
This should be released as it fixes a broken tv guide and livestreams in 2.0.0. Maybe patching 2.0.0 is a better idea?

dagwieers:
I am afraid I already put the v2.2.0 release on Facebook, so people may already have it installed.
That's why I think we need to call it v2.2.1. In itself not a big deal, but v2.2.0 is being delayed as well.

mediaminister:
Okay, no problem.

PR #446, opened 2019-08-16
delete cache when upgrading to version 2.2.1

mediaminister:


resources/lib/vrtplayer.py
```diff
@@ -78,11 +78,11 @@ def _version_check(self):
             # 2.0.0 version: changed plugin:// url interface: show warning that favourites and what-was-watched will break
             if settings_version == '' and self._kodi.credentials_filled_in():
                 self._kodi.show_ok_dialog(self._kodi.localize(30978), self._kodi.localize(30979))
-            if addon_version == '2.2.0':
-                # 2.2.0 version: changed artwork: delete old cached artwork
+            if addon_version == '2.2.1':
```
dagwieers: Shouldn't we be checking 'anything equal or newer than 2.2.0' for a few releases instead?
(I'll merge this, but maybe we can improve this in the master branch after v2.2.1)

Issue #447, opened 2019-08-19
script.module.routing does not support trailing slashes


dagwieers:
### Describe the bug ###
This issue is very layered:
- Using Kodi built-in `ActivateWindow` doesn't handle URLs without a trailing slash right
- And **script.module.routing** does not handle URLs with trailing slashes, unless you explicitly add such routes
- Adding a route for trailing slashes causes trailing slashes for generated URLs

See https://github.com/pietje666/plugin.video.vrt.nu/pull/425#discussion_r313391725

dagwieers:
I opened an issue for this: https://github.com/tamland/kodi-plugin-routing/issues/12

dagwieers:
The fix as been accepted upstream. We are no waiting for a new release.

dagwieers:
The fix has been released!

dagwieers:
The fix we needed from v0.2.2 was reverted in v0.2.3.
Our only recourse is preprocessing what we pass to routing as specified in https://github.com/tamland/kodi-plugin-routing/pull/22#issuecomment-529144310 :-(

PR #449, opened 2019-08-20
Go to 'allseasons'

dagwieers:
I think it is better to jump to a program showing all episodes of all
seasons. This also fixes the issue of the trailing slash.

This relates to #447


mediaminister:
Looks good to me

PR #452, opened 2019-08-22
Add InputStream Helper to Troubleshooting settings

dagwieers:
This PR includes:
- Add InputStream Helper to troubleshooting settings
- Rename "No transmission" to "Retransmissions" (impacts Eén and Canvas)


dagwieers:
Since this is going to happen now anyway, we might as well add it to our plugin.
(Helps in testing too ;-))

PR #453, opened 2019-08-27
Add a screenshot to the README.md

dagwieers:
It appears that when searching for repositories, having a screenshot in
your README will make that appear in the search-results.

So we need a screenshot ;-)


mediaminister:
Ready to merge.

PR #454, opened 2019-08-27
Rename 'My programs' to 'My favorites'

dagwieers:
Since the VRT NU app now also list "Favorieten" we want to make our
naming consisent to the VRT NU app.


mediaminister:
Okay te merge.

PR #458, opened 2019-08-31
Add coverage support

dagwieers:
This adds coverage support to our tests.
It should give a good indication what we are actually testing.


dagwieers:
Looking good, although I would like to stop that codecov-io bot.

dagwieers:
Coverage would be 77% if we enabled authentication (and favorites) support in our Travis tests.

PR #464, opened 2019-09-01
Disable codecov-io github bot

dagwieers:
Attempt to disable @codecov-io


dagwieers:
It seems to work, and also ignoring the test/ directory works this way.

PR #470, opened 2019-09-02
Another attempt to fix Travis

dagwieers:
Still not there with Travis and protected environment variables and
utf-8 output.


dagwieers:
This fixed it. Somehow because we enabled protected environment variables, the encoding for I/O went from utf-8 to ascii, causing a lot of printed testing output to fail. Only when the commit entered the master branch (so this wasn't happening on the PR itself).

PR #474, opened 2019-09-02
Some more testing tweaks

dagwieers:


dagwieers:
It is interesting to see the difference between:
- **Without credentials**:
https://codecov.io/gh/pietje666/plugin.video.vrt.nu/tree/060b68c258e73e96b77129db25d2eae04a352579/resources/lib
- **With credentials**:
https://codecov.io/gh/pietje666/plugin.video.vrt.nu/tree/22febec051fa67b607d5495e1de1e96c2290b02d/resources/lib

dagwieers:
Ok, so I succeeded in not getting these codecov/project errors on checks.
I think this will prove to be more stable and less distracting.

PR #477, opened 2019-09-03
Move cursor down when ListItem is unfollowed/removed

dagwieers:
This fixes an annoying issue where your cursor would jump to the start of
the list when you unfollow a program in the favorite listing, and so the
selected ListItem disappears from under the cursor.

This also implements a smarter way to know where we are within the
add-on at any time.

This requires tamland/kodi-plugin-routing#20

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -9,7 +9,7 @@
         <!-- import addon="inputstream.adaptive" version="2.0.20" optional="true"/ -->
         <import addon="script.module.inputstreamhelper" version="0.3.5"/>
         <import addon="script.module.pysocks" version="1.6.8" optional="true"/>
-        <import addon="script.module.routing" version="0.2.0"/>
+        <import addon="script.module.routing" version="0.2.2"/>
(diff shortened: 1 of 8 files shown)


michaelarnauts:
Interesting "hack", but isn't moving up a better way?

When you remove the last item, you will end on the previous item instead, what will be the next last item.

Not sure what happens now when you are on the last or the first item. Does it loop?

dagwieers:
@michaelarnauts There are two corner-cases (1) you are on the last item (2) the next item is also remove as part of unfollowing a program. In both cases you will jump to the top, and there's not a lot we can do about it. But at least we have fix 90% of the cases with this.

dagwieers:
Still waiting for the [routing PR](https://github.com/tamland/kodi-plugin-routing/pull/20) to be accepted and released.

dagwieers:
The fix for routing (tamland/kodi-plugin-routing#20) has been released.
So this is ready to be merged.

PR #479, opened 2019-09-03
Another test using the proxy variables

dagwieers:
Since I don't see it working, add more output statements.


dagwieers:
Need to be cleaned up later.

PR #486, opened 2019-09-07
support user paths with unicode characters

mediaminister:
Fixes for Björn and Michèle


michaelarnauts:
Nice. You won't believe how much software is broken when you have a non-ascii character in your name.

Issue #361, opened 2019-07-05
Chorus web interface does not support formatting in labels


dagwieers:
### Describe the bug ###
The Chorus web interface does not support text formatting in labels.

Since we added [B] (bold) to "Most recent" listing, Kore shows these tags as-is.

We need to report/fix this upstream.

Related to #249

dagwieers:
Reported as https://github.com/xbmc/chorus2/issues/367

dagwieers:
This seems to have been fixed (by removing the formatting/markup altogether).
Although I cannot find any related commit that implements this.
So not sure if the original report was about something else that I have not reproduced yet :-/

dagwieers:
It would be nice if **bold** and *italic* were supported though (colors I can understand that this is not supported).

Issue #249, opened 2019-05-17
Kore remote control does not support formatting in labels


dagwieers:
### Describe the bug ###
The Kore remote control app for Android does not support text formatting in labels.

Since we added [B] (bold) to "Most recent" listing, Kore shows these tags as-is.

We need to report/fix this upstream.

dagwieers:
Relates issues:
- https://github.com/xbmc/Kore/issues/629
- https://github.com/xbmc/Kore/issues/426

dagwieers:
The Yatse remote control app is not affected by this issue.

dagwieers:
I believe this will be fixed if xbmc/Kore#667 is merged.

mediaminister:
It is merged.

dagwieers:
Indeed, waiting for it to become released though, so we can finally close this.

dagwieers:
I tested the latest beta version of Kore, and unfortunately the title-formatting is not being stripped. So the VRT NU interface still doesn't look very nice in Kore :-( In Yatse things do look nice though.

SyncedSynapse:
xbmc/Kore#688 should fix it

dagwieers:
@SyncedSynapse This is indeed now fixed. There is a small issue that for the thumbnails it is still taking the first character **[** rather than the first character of the sanitized title, but that's just a minor detail.

A working version is available from: http://mirrors.kodi.tv/tools/kore/nightlies/

dagwieers:
So the latest update on Android got the fixes in. We can close this tracking issue.

PR #497, opened 2019-09-19
Prepare for v2.2.3 release

dagwieers:

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,13 @@ leave a message at [our Facebook page](https://facebook.com/kodivrtnu/).
 ![](./resources/media/screenshot04.jpg)
 
 ## Releases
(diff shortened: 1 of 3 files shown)


dagwieers:
There is no real urgency to release, and no real bug-fixes, just minor improvements.
I don't think we need a major release for this.

dagwieers:
However I would like so see the issues related text formatting in Kore (#249) and Chorus (#361) fixed.

mediaminister:
If we mention support for unicode Windows profile paths, we should also make a release for inputstreamhelper because 0.4.2 doesn't support this yet.

dagwieers:
@mediaminister Good point. I think we are indeed ready to also make a release for **inputstreamhelper**, but that is more your department ;-)

PR #385, opened 2019-07-15
Attempt to provide sensible errors on network issues

dagwieers:
Currently the add-on works fine if all is well, however various things
can go wrong which makes the add-on bail out with no proper error.

And since the cause of network errors could be anywhere, we do not want
the end-user to have to dig deep into Kodi logs to find probable
suspects.

So we do want to return the exact error message (i.e. Connection
refused, No route to host, etc.) so the user can relate it to other
possible problems. (WIFI issues, Internet down or unreliable, etc.)

This PR relies on #384 and is far from finished, there are many places in the code that may be affected by network issues. So we have to make sure we are doing this as efficient as possible so the code remains readable and transparent.

PR #500, opened 2019-09-19
Add proxy-support testing

dagwieers:
This is a backport of the proxy-support tests from inputstreamhelper.


dagwieers:
Just as with **inputstreamhelper** on Python 2 the proxy is not as reliable as we would expect, and retrying is required. This would be something I would like to implement at some point (#385) to make the add-on more reliable, but not for now.

PR #506, opened 2019-09-26
Implement support for resumepoints

### LINKED ISSUE #439: Implement resume points and "Watch later" list
The VRT NU website allows to tag programs to watch later and store resume points. This is a useful addition to following programs. It would be nice if our users could use the same facility to tag content to watch later or continue watching on any device.

dagwieers:
This PR includes:
- Implementation for resumepoints and caching
- Add support for Watch later
- Add support for resuming episodes
- Push position when stream is paused/stopped
- Implement Watch Later list

Still to do at some point when Kodi supports it:
- Support 'Mark as unwatched' and 'Reset resume position'

This fixes #439

### DIFF
--- a/resources/lib/resumepoints.py
+++ b/resources/lib/resumepoints.py
@@ -97,7 +97,7 @@ class ResumePoints:
             self._kodi.show_notification(message=self._kodi.localize(30976))
             return False
         # NOTE: Updates to resumepoints take a longer time to take effect, so we keep our own cache and use it
-        self._resumepoints[uuid]['value'] = payload
+        self._resumepoints[uuid] = dict(value=payload)
         self._kodi.update_cache('resume_points.json', self._resumepoints)
         self.invalidate_caches()
         return True
@@ -151,10 +151,8 @@ class ResumePoints:
 
     def invalidate_caches(self):
         ''' Invalidate caches that rely on resumepoints '''
-        # NOTE: Do not invalidate resume_points cache as we manage it ourselves
-        # self._kodi.invalidate_caches('resume_points.json')
-#        self._kodi.invalidate_caches('my-offline-*.json')
-#        self._kodi.invalidate_caches('my-recent-*.json')
+        self._kodi.invalidate_caches('resume_points.json')
+        self._kodi.invalidate_caches('watchlater-*.json')
 
     def refresh_resumepoints(self):
         ''' External API call to refresh resumepoints, used in Troubleshooting section '''
--- a/resources/lib/resumepoints.py
+++ b/resources/lib/resumepoints.py
@@ -80,8 +80,8 @@ class ResumePoints:
 
         if uuid in self._resumepoints:
             payload = self._resumepoints[uuid]['value']
-        else:
-            payload = dict(position=position, total=total, url=url)
+
+        payload = dict(position=position, total=total, url=url)
 
         if watch_later is not None:
             payload['watchLater'] = watch_later
--- a/resources/lib/resumepoints.py
+++ b/resources/lib/resumepoints.py
@@ -55,17 +55,20 @@ class ResumePoints:
         return self._resumepoints
 
     def set_resumepoint(self, uuid, title, url, watch_later=None, position=0, total=100):
-        ''' Set a program as watchlater, and update local copy '''
+        ''' Set program resumepoint or watchLater status and update local copy '''
 
+        # Load current resumepoints
         self.get_resumepoints(ttl=60 * 60)
+
         if watch_later is not None and not position and watch_later is self.is_watchlater(uuid):
-            # Already followed/unfollowed, nothing to do
+            # watchLater status is not changed, nothing to do
             return True
 
         if watch_later is None and position == self.get_position(uuid):
-            # Already followed/unfollowed, nothing to do
+            # resumepoint is not changed, nothing to do
             return True
 
+        # Collect header info for POST Request
         from tokenresolver import TokenResolver
         xvrttoken = TokenResolver(self._kodi).get_xvrttoken(token_variant='user')
         if xvrttoken is None:
@@ -80,15 +83,25 @@ class ResumePoints:
         }
 
         if uuid in self._resumepoints:
+            # Update existing resumepoint values
             payload = self._resumepoints[uuid]['value']
-
-        payload = dict(position=position, total=total, url=url)
+            payload['url'] = url
+            payload['position'] = position
+            payload['total'] = total
+        else:
+            # Create new resumepoint values
+            payload = dict(position=position, total=total, url=url)
 
         if watch_later is not None:
+            # Add watchLater status to payload
             payload['watchLater'] = watch_later
 
+            # Updating watchLater status, invalidate watchlater menu cache
+            self._kodi.invalidate_caches('watchlater-*.json')
+
         import json
-        data = json.dumps(payload).encode('utf-8')
+        from statichelper import from_unicode
+        data = from_unicode(json.dumps(payload))
         self._kodi.log('URL post: https://video-user-data.vrt.be/resume_points/{uuid}', 'Verbose', uuid=uuid)
         req = Request('https://video-user-data.vrt.be/resume_points/%s' % uuid, data=data, headers=headers)
         result = urlopen(req)
@@ -96,10 +109,11 @@ class ResumePoints:
             self._kodi.log_error("Failed to (un)watch episode' at VRT NU")
             self._kodi.show_notification(message=self._kodi.localize(30976))
             return False
+
         # NOTE: Updates to resumepoints take a longer time to take effect, so we keep our own cache and use it
         self._resumepoints[uuid] = dict(value=payload)
         self._kodi.update_cache('resume_points.json', self._resumepoints)
-        self.invalidate_caches()
+        self._kodi.container_refresh()
         return True
 
     def is_watchlater(self, uuid):
@@ -148,11 +162,6 @@ class ResumePoints:
     def resumepoints_uuids(self):
         ''' Return all resumepoints uuids and their resume point '''
         return {key: value.get('value').get('position') for key, value in list(self._resumepoints.items()) if value.get('value').get('position', 0) != 0}
-
-    def invalidate_caches(self):
-        ''' Invalidate caches that rely on resumepoints '''
-        self._kodi.invalidate_caches('resume_points.json')
-        self._kodi.invalidate_caches('watchlater-*.json')
 
     def refresh_resumepoints(self):
         ''' External API call to refresh resumepoints, used in Troubleshooting section '''
--- a/resources/lib/resumepoints.py
+++ b/resources/lib/resumepoints.py
@@ -1,0 +1,184 @@
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2019, Dag Wieers (@dagwieers) <dag@wieers.com>
+# GNU General Public License v3.0 (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+''' Implementation of ResumePoints class '''
+
+from __future__ import absolute_import, division, unicode_literals
+
+try:  # Python 3
+    from urllib.request import build_opener, install_opener, ProxyHandler, Request, urlopen
+except ImportError:  # Python 2
+    from urllib2 import build_opener, install_opener, ProxyHandler, Request, urlopen
+
+
+class ResumePoints:
+    ''' Track, cache and manage VRT resume points and watch list '''
+
+    def __init__(self, _kodi):
+        ''' Initialize resumepoints, relies on XBMC vfs and a special VRT token '''
+        self._kodi = _kodi
+        self._proxies = _kodi.get_proxies()
+        install_opener(build_opener(ProxyHandler(self._proxies)))
+        # This is our internal representation
+        self._resumepoints = dict()
+
+    def is_activated(self):
+        ''' Is resumepoints activated in the menu and do we have credentials ? '''
+        return self._kodi.get_setting('useresumepoints') == 'true' and self._kodi.credentials_filled_in()
+
+    def get_resumepoints(self, ttl=None):
+        ''' Get a cached copy or a newer resumepoints from VRT, or fall back to a cached file '''
+        if not self.is_activated():
+            return
+        resumepoints_json = self._kodi.get_cache('resume_points.json', ttl)
+        if not resumepoints_json:
+            from tokenresolver import TokenResolver
+            xvrttoken = TokenResolver(self._kodi).get_xvrttoken(token_variant='user')
+            if xvrttoken:
+                headers = {
+                    'authorization': 'Bearer ' + xvrttoken,
+                    'content-type': 'application/json',
+                    'Referer': 'https://www.vrt.be/vrtnu',
+                }
+                req = Request('https://video-user-data.vrt.be/resume_points', headers=headers)
+                self._kodi.log('URL post: https://video-user-data.vrt.be/resume_points', 'Verbose')
+                import json
+                try:
+                    resumepoints_json = json.load(urlopen(req))
+                except Exception:  # pylint: disable=broad-except
+                    # Force resumepoints from cache
+                    resumepoints_json = self._kodi.get_cache('resume_points.json', ttl=None)
+                else:
+                    self._kodi.update_cache('resume_points.json', resumepoints_json)
+        if resumepoints_json:
+            self._resumepoints = resumepoints_json
+
+    def set_resumepoint(self, uuid, title, url, watch_later=None, position=0, total=100):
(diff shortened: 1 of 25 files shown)


dagwieers:
This implementation is not yet functional. There's also a technical issue listing the Watch Later list as it only includes assetpath-uuid and url, no title or anything similar. This means we have to do a per-item lookup in order to translate it to something useful.

The resume points however are not affected by this issue.

dagwieers:
There seems to be a bug in **routing**, causing the context-menu to fail to add/remove from Watch Later list.

mediaminister:
> There seems to be a bug in routing, causing the context-menu to fail to add/remove from Watch Later list.

Indeed, the pattern `/watchlater/<uuid>/<title>/<path:url>` is not recognized correctly, changing it to `/watchlater/<path:url>/<uuid>/<title>` works but there are other bugs in resumepoints.py:
- fails if local http caching is disabled in settings
- succeeds if local http caching is enabled but doesn't show up in "Watch later" on VRT NU website

dagwieers:
Oh, I know, that's why it is still a draft.

dagwieers:
Another item I am struggling with is sending position to VRT NU. It seems I have to create my own Player class to get to events on playback (to store video info) and on paused/stopped, but I don't see how we use this Player class (as we're not using it now).

What does work already is continuing a video if it was paused/stopped on the VRT NU website.

mediaminister:
> Another item I am struggling with is sending position to VRT NU. It seems I have to create my own Player class to get to events on playback (to store video info) and on paused/stopped, but I don't see how we use this Player class (as we're not using it now).

Our own Player class can easily be added to our VrtMonitor class in service.py

I implemented a test that shows the stream position when the user pauses, resumes or stops a stream.

https://github.com/mediaminister/plugin.video.vrt.nu/tree/playerposition

dagwieers:
@mediaminister I looked into that, but from the service it is not very straightforward to get to all the information that is required to update this in resume_points. (asset_uuid and url, but also the title for the notification). Feel free to continue on this branch if you like.

mediaminister:
I don't see any other option than using service.py and running a Thread to get Kodi player position.
I doubt we can get resume points from the Kodi database because of this bug: https://github.com/xbmc/xbmc/issues/15897

resources/lib/resumepoints.py
```diff
@@ -84,7 +84,7 @@ def set_resumepoint(self, uuid, title, url, watch_later=None, position=0, total=
             payload = dict(position=position, total=total, url=url)
 
         if watch_later is not None:
-            payload['watchLater'] = watch_later
+            payload['watchLater'] = str(watch_later).lower()
```
dagwieers: Are you sure this is correct? The JSON payload we get back from: https://video-user-data.vrt.be/resume_points are proper JSON booleans, not strings.

dagwieers: I can also confirm that in the request, true booleans are being used.

mediaminister: I'm not sure, I will remove this again.

dagwieers:
BTW We cannot merge this before we release v2.2.3. I am waiting for the **inputstreamhelper** release for this.

dagwieers:
Maybe you already tested this, but I tried using:
```python
self.log('ResumeTime: %s' % play_item.getProperty('ResumeTime'))
```

But it always returns `0.000000`.

dagwieers:
I think this should work:
```python
        player = xbmc.Player()
        while not player.isPlaying() and not xbmc.Monitor().abortRequested():
            xbmc.sleep(100)
        player.showSubtitles(subtitles_visible)

        while player.isPlaying():
            try:
                resumetime = player.getTime()
            except RunTimeError:
                break
            xbmc.sleep(1000)

        from resumepoints import ResumePoints
        _resumepoints = ResumePoints(self)
        ...
```

Although we currently do not have the required information available at play time.
Two options:
- We pass it through as additional options
- We perform a lookup (either when playing, or when stopping)

mediaminister:
> we currently do not have the required information available at play time.

That's the problem where I'm stuck. I can expose the timestamps from Kodi player in the kodiwrapper `play` function and I can get uuid/url/title from all listitems in kodiwrapper `show_listing`.
But I didn't find an acceptable method yet to get this info from the selected listitem.
I don't like editing the `plugin:// ` url api for this or requesting the same info from VRT search API twice.

I'm currently looking in `xbmcgui`  class and JSON-RPC methods to find a way.

I also found out that VRT NU caches resumepoints in the webbrowser with the [Indexed DB API ](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API)
This implies that an updated resumepoint from Kodi to VRT NU doesn't work instantly. The `preference-store-cache` which contains key/value pairs for the resumepoints has to be updated first. Because the updated resumepoints come from Kodi, the VRT NU website doesn't know the cache should be updated. Maybe we need a webbrowser add-on to  support resumepoints from Kodi

dagwieers:
We can always drop this feature if it is too complex to add. But the "Watch list" is also not easy to get without adding a lot of lookups. Maybe we should document why this is hard to do and may not be worth the trouble?

dagwieers:
I am going to move the non-resumepoints changes out of this PR, and move them in a separate PR.

mediaminister:
It's not too complex, I got this already working by abusing a property from `xbmc.getInfoLabel` and `xbmc.Player().getVideoInfoTag()`
It's just a matter of finding the right api to do this and comply with Kodi rules.

> I am going to move the non-resumepoints changes out of this PR, and move them in a separate PR.

Okay,  no problem.

dagwieers:
This means I will rebase this PR after merging those changes. If you have any outstanding stuff that belongs to this PR, it is probably easier to commit it now.

mediaminister:
Go ahead, you can also remove my `monitor stream position` commit because I moved this to a separate class and file.

mediaminister:
I implemented _Push position when stream is paused/stopped_

Clearing the `preference-store-cache` in the Indexed DB Storage of your webbrowser is needed to get the updated resumepoint working (only needed when you previously watched a part of the same program in your webbrowser.)

mediaminister:
> There's also a technical issue listing the Watch Later list as it only includes assetpath-uuid and url, no title or anything similar. This means we have to do a per-item lookup in order to translate it to something useful.

I implemented the watch later list, only one lookup was needed:
```
https://vrtnu-api.vrt.be/search?i=video&facets[url]=[//www.vrt.bevandaag/2019/vandaag-d20190926,//www.vrt.be/vrtnu/a-z/het-journaal/2019/het-journaal-het-journaal-19u-20191006/,//www.vrt.bevrtnu/a-z/vandaag/2019/vandaag-d20190926/,//www.vrt.be/vrtnu/a-z/sportweekend/2019/sportweekend-d20191006/]&size=300
```
There are still bugs with caching, unit tests and code needs refactoring.

mediaminister:
I think my last commit got "watch later" and "resumepoints" working without errors in Kodi.

I would like to implement the last missing functions, but I don't understand the difference.

> Support 'Mark as unwatched' and 'Reset resume position'

@dagwieers Can you explain the difference between those two functions?

dagwieers:
@mediaminister Both are already present options in the context menu of Kodi. I don't know if we can intercept their use and act on it. (If not, that is something to have fixed in Kodi).

mediaminister:
I think it's not easy to implement this because the argument `replaceItems` was removed from `xbmcgui.ListItem.addContextMenuItems` There is a discussion about this, also mentioning the use of sending watched status to third party services: https://forum.kodi.tv/showthread.php?tid=270648&pid=2430500#pid2430500

dagwieers:
I opened a forum thread wrt. general Video-on-Demand support in Kodi. I think VOD plugins could be handled a lot better with some support from Kodi (and eventually skins).
https://forum.kodi.tv/showthread.php?tid=348138

Feel free to add any thoughts or opinions, as I think some wider feedback would be useful to come to a better understanding altogether.

dagwieers:
This is almost finished, I noticed the following possible improvements:
- Show notifications when a video is added/removed from watchlater list
- The "My watch later" listing does not indicate correctly the programs that are followed
- The currently played ListItem is not updated with the newer resumepoint (or the virtual listing is not refreshed) so restarting a stopped video does not resume correctly. Refresh fixes it.

mediaminister:
> - Show notifications when a video is added/removed from watchlater list

Is this not implemented yet?

> - The "My watch later" listing does not indicate correctly the programs that are followed

Fixed in https://github.com/pietje666/plugin.video.vrt.nu/pull/506/commits/87f40baef11e1e3eb1c3ba283daf1f0bae116015

> - The currently played ListItem is not updated with the newer resumepoint (or the virtual listing is not refreshed) so restarting a stopped video does not resume correctly. Refresh fixes it.

I noticed that in the master branch Kodi automatically updates the internal resume points when playback is stopped.

With our custom resume points implementation, it seems Kodi doesn't automatically updates the internal resume points anymore. I should investigate why this is the case.

I provisionally fixed it with a call to `container_refresh` after setting a resumepoint in https://github.com/pietje666/plugin.video.vrt.nu/pull/506/commits/a459918a553692ceed688eb1b88ec7c7411a3ac6

dagwieers:
> Is this not implemented already?

Let me test again, I think I did not see them yesterday. But did not have time to look into it to fix it.

> With our custom resume points implementation, it seems Kodi doesn't automatically updates the internal resume points anymore. I should investigate why this is the case.

I think the problem is that we have set that information in the ListItem which overrides that internal position. If you explicitly refresh, this will override the position with the correct resumepoint. Maybe disabling caching for this virtual directory is enough? (Caching was enabled)

Let me test this as well.

dagwieers:
So, disabling caching does not really help, but is probably better anyway. (we do this for favorites as well).

There is one more thing I we should probably add as well. Have a list of programs you have not finished yet. In Netflix this is called a "Continue watching" list. Should be pretty easy to add based on what you did for the "Watch list".

dagwieers:
All in all, this is working pretty well. I am happy to merge this and release it as part of v2.3.0.

mediaminister:
> There is one more thing I we should probably add as well. Have a list of programs you have not finished yet. In Netflix this is called a "Continue watching" list. Should be pretty easy to add based on what you did for the "Watch list".

It seems VRT NU filters out some programs from the resumepoints json and only displays a limited number of programs in the "Programma's verder kijken"  list.
In my "Continue watching" implementation I added a similar filter.

dagwieers:
The "Continue watching" listing is empty here, while I have 3 items on my Watch list that were not finished. So I guess that it what is being filtered. Personally I would not filter it out. I understand why they are doing it on the website (screen estate is valuable and showing the same stuff both in the "To watch" list and in the "Continue watching" list does not add any value.

But in our case, these two lists do not appear on the screen at the same time, so I do think we should provide an unfiltered list (or at least make this behaviour optional).

dagwieers:
Also, what irritates me the past few weeks is that the "My favorites" lists a lot of entries all starting with "My". The reason why we started doing this was because we did not want to confuse the user between the "Main menu", and the "My favorites" listing.

But I don't think this is longer an issue.
1. We now have breadcrumbs
2. The listing looks sufficiently different (not just in size, but also in entries)

I would like to improve this:
- Use proper breadcrumbs for the submenus (My programs, My most recent, My soon offline, etc...)
- Maybe use a different artwork, or change the default colours of the menu, to indicate a submenu

And then we can simply have:
- My programs
- Most recent
- Soon offline
- Continue watching
- Watch later
- All movies
- One-off documentaries

dagwieers:
> I agree. Maybe we should also rename "My favorites" because "watch later" and "continue watching" is not really related to favorites. In that case some plugin:// urls should also be changed.

That's a fair point from an implementation perspective, but shouldn't matter to the end-user. Something they have watched, or want to watch later could also be considered "favorites".

So I don't think that needs a change per se, but I don't have strong feelings to the "My favorites" title either, a better proposal is welcome. I do think logically this belongs together in a single submenu (and I am so used to it that I spend 95% in the My favorites submenu, and only frequent the main "Recently added" menu to see if there's something I should be following).

From a routing perspective, I don't mind if it follows the implementation, rather than a logical structure. There's a lot to say for that actually. For resumepoints we have to decide now :-)

mediaminister:
Okay, leave it to "My favorites".

dagwieers:
Regarding the context menu with "Mark as watched", "Marks as unwatched" and "Reset resume position", I noticed that Netflix does not have any of these entries. So what we could be doing is remove these entries from the context menu, and add our own versions of these.

mediaminister:
> So what we could be doing is remove these entries from the context menu, and add our own versions of these.

I wonder how Netflix add-on is doing this, I thought it was nearly impossible. I'll look into this.

I edited the 'continue watching' filter according to VRT NU's implementation: programs that are only played for the first 30 seconds or stopped the last 30 seconds of the program are filtered out. I removed the date filter.

I added an "**Enable watching activity**" setting entry to the add-on settings. I think this label represents all resumepoint related functionality in a comprehensible term.

dagwieers:
I completely agree !

mediaminister:
> Regarding the context menu with "Mark as watched", "Marks as unwatched" and "Reset resume position", I noticed that Netflix does not have any of these entries.

I checked the Netflix add-on, these context menu items show up as soon as you have partially watched an item. And more important: Netflix add-on doesn't sync Kodi player position to the Netflix website and vice versa.

dagwieers:
@mediaminister You are right, the confusion happened because it is not always clear what is a tvshow and what is an episode in the Netflix add-on. So I guess I was testing it on the wrong entries. I do see a "Mark as watched".

We will have to get this "fixed" in Kodi one way or another.

dagwieers:
Let's merge this. Thanks for this major piece of work!

dagwieers:
I discovered that older videos do not have `assetPath` and as such do not have this **Watch later** context menu item. The add-on handles this correctly though.

The **Continue watching** listing is still empty for me, so I still need to investigate what is going on there.

dagwieers:
Another strange visual quirk happens when after we have played a video, and we refresh the container it suddenly goes from 'tvshow' content to 'video' content (showing the thumbnail as icons).

mediaminister:
> The Continue watching listing is still empty for me, so I still need to investigate what is going on there.

This happens when an empty "Continue watching" menu is cached. I think it is not automatically refreshed.

> Another strange visual quirk happens when after we have played a video, and we refresh the container it suddenly goes from 'tvshow' content to 'video' content (showing the thumbnail as icons).

I saw this once, but I can't remember when and in which menu.

dagwieers:
So after investigating, it appears the query we perform doesn't match every entry properly. The request is this one:

- https://vrtnu-api.vrt.be/search?i=video&facets[url]=[//www.vrt.be/vrtnu/a-z/professor-t/3/professor-t--s3a3/,//www.vrt.be/vrtnu/a-z/eigen-kweek/2/coloradokevers/,//www.vrt.be/vrtnu/a-z/studio-ibiza/1/studio-ibiza-s1a1-genesis/,//www.vrt.be/vrtnu/a-z/bob-de-bouwer/seizoen-19/bob-de-bouwer-sseizoen-19a254-bob-de-bouwer-r019-a0254/,//www.vrt.be/vrtnu/a-z/belpop-classics/2/belpop-s2a5/,//www.vrt.be/vrtnu/a-z/de-afspraak-op-vrijdag/2019/de-afspraak-op-vrijdag-d20190510/#,//www.vrt.be/vrtnu/a-z/de-campus-cup/1/campus-cup-s1a22/,//www.vrt.be/vrtnu/a-z/de-zevende-dag/2018/de-zevende-dag-d20180211.ch.b20b9461-2621-431c-a1a1-ae733eb42bea/]&from=1&size=50

And it contains 2 resumepoints of interest that do not return a match:
- //www.vrt.be/vrtnu/a-z/belpop-classics/2/belpop-s2a5/
- //www.vrt.be/vrtnu/a-z/studio-ibiza/1/studio-ibiza-s1a1-genesis/

dagwieers:
If I query for this:
- https://vrtnu-api.vrt.be/search?i=video&facets[url]=[//www.vrt.be/vrtnu/a-z/belpop-classics/2/belpop-s2a5/]&from=1&size=50

It does work, so something else is wrong here.

dagwieers:
Another weird issue, while testing I watched the latest De Ideale Wereld episode and scrolled forward (position 05:05) and stopped playing. Then I refreshed caches and when into "Continue watching" menu and *all* my previous "Continue watching" entries (I think everything in my resume_points.json that is current) suddenly was set to position 05:05.

dagwieers:
Ok, I can no longer reproduce this after rebooting Kodi. My assumption is that there was something wrong with the Player() object. Will have to test again when Kodi has been running for a longer time.

dagwieers:
When I stop playing midway, I count 5 different calles to resumepoints for the episode played. And a massive amount of deletes and refreshes. So there is something off :-/

PR #507, opened 2019-09-29
Various fixes

dagwieers:
This includes changes from the resumepoints PR that are not specific to
resumepoints.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,7 +84,8 @@ msbuild.wrn
 kodi-addon-checker-report.log
 kodi-addon-checker.log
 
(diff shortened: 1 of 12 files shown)


dagwieers:
I included the `get_properties()` functionality in *metadata.py*, even though it is not yet used.
Not sure if that is a good idea, both the `episodeno` property as the `year` property are not making a real difference (as I had hoped). They come from other plugins.

I added the `IsInternetStream` property here, as it does seem to be used (by 6 different add-ons). Not sure what the impact is of setting this.

PR #511, opened 2019-10-06
Add colour preferences to accommodate light skins

### LINKED ISSUE #508: Use of colors can make the addon hard to use on lighter skins.
# Describe the bug ###
The use of colors to highlight some things should be avoided since light skins make these strings hard to read.

image

The color of the geoblocked and the "beschikbaar tot" is readable on my skin, but not so pretty. I can imagine it also being hard to see on other skins.

image

dagwieers:
This allows to set the colouring theme for certain text used in the
plot or in listings.

This fixes #508


dagwieers:
@michaelarnauts Please test the light-mode and custom colours.

We may want to rethink the different colours offered (maybe include light/dark variants as well).
https://github.com/xbmc/xbmc/blob/master/system/colors.xml

dagwieers:
@pietje666 @mediaminister What do you think? Is this a problem that needs fixing, and is this the right fix?

mediaminister:
I think you should at least use the BBCODE variable syntax if you want to use variables:
```
[COLOR={highlighted}]{text}[/color]
```
And then you can format these strings like this:
```
'[COLOR={greyedout}][I](for protected content)[/I][/COLOR]'.format(greyedout='darkgray')
```

dagwieers:
@mediaminister That is more problematic, because it requires us to know in advance the values for each of these colors. (In effect, thanks to our custom dicts, we can still replace by string, but I guess that is not what you would like to do anyway). Let me think this through.

dagwieers:
Also, according to the documentation the syntax in Kodi is `[COLOR color]` and not `[COLOR=color]`.
https://kodi.wiki/view/Label_Formatting

But I have seen both being used in plugins, although the syntax without equal sign is more prevalent.

mediaminister:
Both are bbcode, the one with the equal sign is the newer syntax. Looking at the revision history of "Label Formatting" Kodi wiki, I think this page needs an update to the newer syntax.

>  That is more problematic, because it requires us to know in advance the values for each of these colors.

I don't understand why. I only made a code syntax remark.I just want it to be clearly visible in the codebase that we are using variable colors.

Does the following code changes your logic?

```
'[COLOR={greyedout}][I](for protected content)[/I][/COLOR]'.format(greyedout=self._kodi._get_color('greyedout'))
```

pietje666:
Its cool that users can change the colours, the addon really looks cool these days btw 👍

michaelarnauts:
@dagwieers these colors looks a lot better on my light skin!

dagwieers:
Now that we have finished the big changes for resumepoints, I can focus a bit more on this one.

mediaminister:
@dagwieers I implemented the BBCODE variable syntax. Can you take a look at this again?

dagwieers:
The implementation looks fine, but because you force-pushed to my branch, I cannot update it or I loose my original changes. :-/ A new commit in this case would have been more helpful to compare.

dagwieers:
I am going to merge this. Tests worked out fine.

PR #514, opened 2019-10-08
Add availability and Geoblocked on the same line

dagwieers:
If it fits, this adds Geo-blocked on the same line as availability.

This works in English and not in Dutch (based on screen-size).
Since the colours give a visual separation, we don't need to add a
separator (which would ruin this).

I much rather added a Lock picture/character/emoji, but that is not
possible at this time.

The main reason for doing this is to save a bit on screen-estate, both in the plot-section as well as in the Information pane (which has fewer vertical space).

This PR also added/changed a few comments for localize().


michaelarnauts:
Actually, a question. Is it possible to detect if you are roaming based on an API result? Maybe it makes sense to hide the geoblocked if you can detect that you can play it anyway. I don't care about this information since I only use Kodi at home.

dagwieers:
IMO this should be part of the video properties at the bottom-right with all other metadata and it could be a simple lock-icon (or a globe + lock icon) in Estuary skin.

And also a property in the Information pane.

I opened a thread related to all Video-on-Demand specific needs in Kodi here:
https://forum.kodi.tv/showthread.php?tid=348138

dagwieers:
I am going to merge this so I can test this with resumepoints, if there are objections or improvements we can add those later.

Issue #299, opened 2019-06-08
Some tvshows fail to work with favorites


dagwieers:
### Describe the bug ###
We match the TV-shows  `programName` with the favorites name, but this is apparently not very standardized as they may be different, e.g. favorites `de-campus-cup` has programName `campus-cup`.

mediaminister:
Indeed, `programName` is unreliable. I noticed this also with the `get_latest_episode` function.
We should always use `programUrl`:
```
//www.vrt.be/vrtnu/a-z/de-campus-cup/
```
The Favorites video-data api uses a slightly shorter `programUrl` format:
```
/vrtnu/a-z/de-campus-cup/
```
And then we have the VRT suggest api that we use to get a-z and category menus. This api uses the ` targetUrl` format:
```
//www.vrt.be/vrtnu/a-z/de-campus-cup.relevant/
```
Luckily, these three `programUrl` variants can be derived from each other. I made a pull request which implements this functionality so the relevant api requests always use `facets[programUrl]` instead of `facets[programName]`: https://github.com/pietje666/plugin.video.vrt.nu/pull/300

dagwieers:
You beat me to it, I was planning to fix this tonight. However I rather not pass programUrl, but would prefer the program-part of the programUrl, and translate to the desired format when used.

mediaminister:
Fixed for [v1.10.1](https://github.com/pietje666/plugin.video.vrt.nu/tree/v1.10.1) in https://github.com/pietje666/plugin.video.vrt.nu/commit/01795435efc91781572660910f60d731169bb4e0

PR #515, opened 2019-10-08
Use programUrl instead of targetUrl

dagwieers:
Not sure why we are using targetUrl in the suggest API while we have a
programUrl (that does not have .relevant).

There probably is a good reason, but at least this gives the opportunity
to discuss it :-)

### DIFF
--- a/resources/lib/apihelper.py
+++ b/resources/lib/apihelper.py
@@ -213,7 +213,7 @@ class ApiHelper:
         oneoff_programs = [statichelper.url_to_program(episode.get('programUrl')) for episode in oneoffs]
 
         for tvshow in tvshows:
-            program = statichelper.url_to_program(tvshow.get('targetUrl'))
+            program = statichelper.url_to_program(tvshow.get('programUrl'))
 
             if use_favorites and program not in favorite_programs:
                 continue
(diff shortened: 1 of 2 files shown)


mediaminister:
If I remember correctly (https://github.com/pietje666/plugin.video.vrt.nu/issues/299#issuecomment-500207230), `programUrl` was not available in the suggest API before.

I tested it and it is safe to use `programUrl` instead of  `targetUrl`.

I also tested `programName` and this still an unreliable variable, you can't convert this to programUrl or targetUrl for the following programs:


|programName|targetUrl|
|---|---|
|de-week-van-karrewiet-vgt|//www.vrt.be/vrtnu/a-z/de-week-van-karrewiet-met-vgt.relevant/|
|f-c--de-kampioenen|//www.vrt.be/vrtnu/a-z/fc-de-kampioenen.relevant/|
|karrewiet-vgt|//www.vrt.be/vrtnu/a-z/karrewiet-met-vgt.relevant/|
|sportweekend-vgt|//www.vrt.be/vrtnu/a-z/sportweekend-met-vgt.relevant/|

dagwieers:
Yeah, that was one of my hypothesis. I was looking and hoping Geo-blocked information was available from the suggests API, but that is not (yet) the case :-/

PR #520, opened 2019-10-10
Various resumepoint fixes

dagwieers:
This PR includes:
- Translation fixes wrt. context menu
- Fix breadcrumbs for Watch later and Continue watching
- Add routing integration tests for Watch later and Continue watching
- Fix issue during testing when resumepoints is not enabled
- Add a 'Refresh watching activity' to Expert settings menu
- Simplify Favorites/ResumePoints interface
- Use total-value from resumepoints for metadata resumetime
- Fix issue causing empty 'Continue watching' listing
- Fix issue with caching of Watch later and Continue watching listings

### DIFF
--- a/resources/lib/playerinfo.py
+++ b/resources/lib/playerinfo.py
@@ -17,20 +17,22 @@ class PlayerInfo(Player):
         self._tracker = None
         self._last_pos = 0
         self._total = None
+        self._stop = False
         Player.__init__(self)
-        while not self._monitor.abortRequested():
-            if self._monitor.waitForAbort(10):
+        while not self._monitor.abortRequested() and not self._stop:
+            if self._monitor.waitForAbort(5):
                 break
 
     def onAVStarted(self):  # pylint: disable=invalid-name
         ''' called when Kodi has a video or audiostream '''
         self._total = self.getTotalTime()
-        self._tracker = Thread(target=self.stream_position)
+        self._tracker = Thread(target=self.stream_position, name='VRTNUThread')
         self._tracker.start()
 
     def onPlayBackStopped(self):  # pylint: disable=invalid-name
         ''' called when user stops Kodi playing a file '''
         self._info((self._last_pos, self._total))
+        self._stop = True
 
     def onAVChange(self):  # pylint: disable=invalid-name
         ''' called when Kodi has a video, audio or subtitle stream. Also happens when the stream changes. '''
@@ -38,9 +40,11 @@ class PlayerInfo(Player):
     def onPlayBackEnded(self):  # pylint: disable=invalid-name
         ''' called when Kodi stops playing a file '''
         self._info((self._total, self._total))
+        self._stop = True
 
     def onPlayBackError(self):  # pylint: disable=invalid-name
         ''' called when playback stops due to an error '''
+        self._info((self._total, self._total))
 
     def onPlayBackPaused(self):  # pylint: disable=invalid-name
         ''' called when user pauses a playing file '''
@@ -53,4 +57,4 @@ class PlayerInfo(Player):
         ''' get latest stream position while playing '''
         while self.isPlaying():
             self._last_pos = self.getTime()
-            sleep(100)
+            sleep(500)
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -371,10 +371,18 @@ msgid "In 2 days"
 msgstr ""
 
 msgctxt "#30401"
-msgid "Add {title} to watch later"
+msgid "Watch later"
 msgstr ""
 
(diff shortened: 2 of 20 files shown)


resources/lib/playerinfo.py
```diff
@@ -17,30 +17,36 @@ def __init__(self, **kwargs):
...
                 break
+        self._info((self._last_pos, self._total))
 
     def onAVStarted(self):  # pylint: disable=invalid-name
         ''' called when Kodi has a video or audiostream '''
         self._total = self.getTotalTime()
         self._tracker = Thread(target=self.stream_position)
+        self._tracker.name = 'VRTNUThread'
         self._tracker.start()
+        self._tracker.name = 'VRTNUThread'
```
mediaminister: Naming this Thread once is enough.

dagwieers: It doesn't work, so I was trying it once before and after the start. This PR is still a work in progress :-/

resources/lib/playerinfo.py
```diff
@@ -17,30 +17,36 @@ def __init__(self, **kwargs):
...
         self._last_pos = 0
         self._total = None
+        self._stop = False
         Player.__init__(self)
-        while not self._monitor.abortRequested():
-            if self._monitor.waitForAbort(10):
+        while not self._monitor.abortRequested() and not self._stop:
+            if self._monitor.waitForAbort(1):
                 break
+        self._info((self._last_pos, self._total))
```
mediaminister: What's the purpose of moving the info update here?

resources/lib/playerinfo.py
mediaminister: Strange...

michaelarnauts:
A question, why always force pushing to the topic branch? When a PR is merged, it can simply by squashed to one commit and merged on master.

Force push makes it impossible to see the changes compared to a previous review. 

Just wondering, maybe I'm missing something here.

dagwieers:
Because for every little test you want to do, you have to commit to get a new versioned zip. And there is no use in having 250 commits with stuff you tried and failed. There is no need to review this PR until it is finished. But GitHub doesn't allow to turn it back into Draft mode, so that sucks.

dagwieers:
It would have been useful to have a few PRs for different changes, but then it becomes a lot harder to test those changes together. Unless you can merge them quickly, but with a review process that would make everything slower.

So in this unusual case (large changes touching lots of stuff for a single feature) there's no simple solution IMO.

resources/lib/playerinfo.py
dagwieers: It was also for testing the behaviour, we should revert this, as it impacts when the container refresh happens. What I discovered is that none of the threads ever stopped, so after playing and stopping 10 videos, the older threads act also on the older videos, so we have to break out of the thread ourselves and that fixes the effect I saw where when I stopped the 10th video, the previous 9 videos would have the same resumetime.

dagwieers: At some point I went nuts, as I couldn't figure out what was happening (because I randomly started videos without recollecting which ones I started so not seeing the theme...).

dagwieers:
This PR also renames some methods for Favorites and ResumePoints which makes thing more clear IMO. And there's a big overhaul in how we invalidate caches, as I noticed files being invalidated, cached an reinvalidated when we had all those threads updating resumepoints and container refreshes in parallel. So while that will not happen in the future again, it's still a lot more efficient now.

dagwieers:
I am not 100% certain we have covered all issues now, resumepoints certainly add unwanted complexity because Kodi does not offer acceptable interfaces for not uncommon VOD operations..

mediaminister:
There exists an acceptable interface that can simplify things, but it not usable because of a bug: https://github.com/xbmc/xbmc/issues/15897
FYI:
Setting resume point to database  based on` plugin://` url
```json
{"jsonrpc": "2.0", "method": "Files.SetFileDetails", "params": {"file": "plugin://plugin.video.test/play/id", "media": "video", "playcount": 2, "resume": {"position": 400.25, "total": 624.87}}, "id": "1"}
```
Getting a resume point from the database based on` plugin://` url
```json
{"jsonrpc": "2.0", "method": "Files.GetFileDetails", "params": {"file": "plugin://plugin.video.test/play/id", "media": "video", "properties": ["resume"]}, "id": "1"}
```


```

```

dagwieers:
@mediaminister Yeah, you mentioned that earlier, but we cannot use it. Even when this is fixed, we have to live with existing Kodi systems that do not work :-/ The current implementation does work well, now that we have fixed the known issues.

My current concern is that the update/refresh takes a lot longer. Need to look if we can improve that somehow.

dagwieers:
So let's merge this and carry over the known issues.

PR #521, opened 2019-10-12
Support for Up Next service add-on

### LINKED ISSUE #420: Add support for "Up Next" Addon
The "Up Next" Kodi addon provides netflix-like binge watching features. When an episode is finished, it can show a message to play the next episode. This works since an addon can emit a message with the information about the next episode.

More information is available here: https://forum.kodi.tv/showthread.php?tid=336747

Information on how to integrate is here: https://github.com/im85288/service.upnext/wiki/Addon-Integration

mediaminister:
This fixes #420

The experimental support includes:
- Functional when Up Next service add-on is installed
- No additional add-on requirements, communication using standard Kodi  `JSONRPC.NotifyAll` API
- Support for continuously playing all episodes of a program across all seasons


resources/language/resource.language.nl_nl/strings.po
```diff
@@ -693,6 +692,14 @@ msgctxt "#30930"
 msgid "This enables 'watch later' and 'continue watching' menus and syncs playback position between different devices using the same VRT NU account."
 msgstr "Dit activeert 'kijk later' en 'verderkijken' menu's and synchroniseert de afspeelpositie tussen verschillende toestellen die hetzelfde VRT NU account gebruiken."
 
+msgctxt "#30931"
+msgid "Enable Up Next (experimental)"
```
dagwieers: I am not sure we want to add this optionally.
I mean if people installed Up Next, it is because they want to use this.
For me that is sufficient as "activation" (provided it works out-of-the-box).
So if we add this (e.g. to advertise this integration) I would enable it by default from the very start.

resources/lib/apihelper.py
```diff
@@ -363,8 +450,8 @@ def get_latest_episode(self, program):
         video = dict(listitem=video_item, video_id=episode.get('videoId'), publication_id=episode.get('publicationId'))
         return video
 
-    def get_episodes(self, program=None, season=None, category=None, feature=None, programtype=None, keywords=None, whatson_id=None, video_id=None,
-                     video_url=None, page=None, use_favorites=False, variety=None, cache_file=None):
+    def get_episodes(self, program=None, season=None, episodes=None, category=None, feature=None, programtype=None, keywords=None,
```
dagwieers: This `get_episodes()` is becoming one big filtering method :-)
It works well and is quite easy to read, but it somehow looks wrong as an interface.
Not sure what the solution is, everything I can think of seems worse.

dagwieers:
There is some overlap with the resumepoints implementation, so maybe we should merge that one PR and carry over any known issues into a new PR. The resumepoints is far from ready to be released. I think it needs at least a month of testing because I don't trust it yet :-/

resources/lib/apihelper.py
mediaminister: I think it's quite normal that we use a lot of parameters when communicating with an extensive API like the VRT Search API. We can combine data using objects, but I didn't take the time yet to think about a better data design. Metadata collection and processing is definitely something that can improve.

mediaminister:
I'm in favor of merging https://github.com/pietje666/plugin.video.vrt.nu/pull/520
I'll test the master branch next week so we can release, if all goes well, by the end of October.

dagwieers:
I just tested this and it works really well.

The settings entry under **My favorites** is a bit weird as there is no relation to Favorites, so I would move it to the **Features** subsection.

We could also split up the current **Interface** tab into:
- My favorites
- Features (with NextUp being part of this, or in a separate tab)

Then it is less crowded. But I still think if the integration works well, there shouldn't be a (visible) option for this.

@michaelarnauts What do other add-ons do in this regard ?

michaelarnauts:
I also think that there should be no option to turn it off. When you install that addon, you want that feature. I think this add-on already contains a lot of settings where the defaults are fine imo.

The add-on is called "Up Next" btw, "Next Up" is an old version without this API.

resources/language/resource.language.nl_nl/strings.po
michaelarnauts: I agree

mediaminister: Dropped

dagwieers:
@mediaminister FYI You can use 'fixes' to automatically link to (and close) the issue, and you can simply type `#420` to reference the issue.

resources/lib/service.py
```diff
@@ -20,6 +21,22 @@ def __init__(self):
             if self.waitForAbort(10):
                 break
 
+    @staticmethod
+    def onNotification(sender, method, data):  # pylint: disable=invalid-name
+        ''' Handler for notifications '''
+        _kodi = KodiWrapper(None)
+        _kodi.log('Got a notification: %s, %s, %s' % (sender, method, to_unicode(data)), 'Verbose')
```
dagwieers: I would prefer this gets removed eventually, as it harms troubleshooting/testing on Kodi (i.e. unrelated notifications).

mediaminister: Okay, when I'll remove the draft status, this will be gone.

dagwieers:
We have been testing it, and it appears to work well.

I also noticed that the Netflix add-on does include in its settings a switch to disable it, and a reference to the Up Next settings. This from the **Playback** tab.

mediaminister:
> I also noticed that the Netflix add-on does include in its settings a switch to disable it, and a reference to the Up Next settings.

if you have no objection I will add a comparable switch because I still think users should be able to disable it when they find "Up Next" an annoying feature.

I'm still not sure if we should only implement this feature for specific program types: `daily, reeksaflopend, oneoff, reeksoplopend`

Detecting credits seems almost impossible because VRT doesn't provide timestamps for credits. VRT provides timestamps for trailers, banners and promotionals, but functionality that uses this data will not be part of this pull request. I'll look into this at a later time.

dagwieers:
@mediaminister Feel free to add the switch as well. I have been flip-flopping over this so I don't expect anyone to care about my opinion at this point ;-)

dagwieers:
Just noticed that for "De Ideale Wereld" the pop-up showed `SNoneE14` and it didn't play the next episode. Possibly the reason for this issue:
```
2019-10-18 15:30:48.891 T:1086641024  NOTICE: [plugin.video.vrt.nu] Got VRT NU Player info: {'position': 2738.778076171875, 'total': 2739.000244140625}
2019-10-18 15:30:48.892 T:1086641024   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                            Error Type: <type 'exceptions.UnboundLocalError'>
                                            Error Contents: local variable 'episode' referenced before assignment
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/resources/lib/playerinfo.py", line 51, in onPlayBackEnded
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/resources/lib/vrtplayer.py", line 356, in <lambda>
                                                video_url = video.get('video_url').replace('https:', '') + '/'
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/resources/lib/vrtplayer.py", line 363, in handle_info
                                                self._resumepoints.update(uuid=uuid, title=title, url=url, watch_later=None, position=position[0], total=position[1])
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/resources/lib/vrtplayer.py", line 382, in push_position
                                            UnboundLocalError: local variable 'episode' referenced before assignment
                                            -->End of Python script error report<--
```

dagwieers:
So it appears I had an older build, so scratch my recently reported issues. They may have all been related to this (can't be sure). Maybe another reason to get this merged ;-)

mediaminister:
This can't be merged yet because there is a problem with threading and the `playerInfo` object.
For every episode a new `playerInfo` object is initialized but it gets never deleted. So, we end up with multiple player objects after watching a couple of episodes.
I'm currently investigating how to fix this properly.

dagwieers:
Aha, that looks like the same problem we had before with resumepoints. And it explains the multiple calls for a single video.

dagwieers:
Could it be onAVChange that needs the PlayerInfo to stop with Up Next?

dagwieers:
And probably onPlayBackError also needs to stop the PlayerInfo object?
Maybe there is a better way to detect when a PlayerInfo object is no longer needed...

mediaminister:
I working on I fix already, I tracked down a couple of problems:
- the player should be stopped from Python before playing a next episode, Up Next add-on doesn't take care of that
- the stream position thread should be stopped immediately when the player is stopped.

The last problem I'm trying to fix is that `onAVStarted` gets called two times. It seems that the previous infoPlayer object gets deleted after a few seconds, but that is actually too late.

dagwieers:
Is this related to: https://github.com/im85288/service.upnext/commit/a39c8daebad5a50f92dd61c93687cb58d1be0077

mediaminister:
I didn't investigate Up Next add-on source code, but there is definitely room for improvement. I'm confident that the remaining problems can be fixed in VRT NU source.

Two remaining problems:
- `onAVStarted` gets called two times sometimes, possibly because the previous infoPlayer object is removed too late (should debug it more to be sure)
- Kodi doesn't mark consecutive episodes as watched when using Up Next feature

mediaminister:
The threading problem is fixed! It turned out we just can use the `isPlaying()` method to exit the while loop when Kodi player stops and decreasing the `waitForAbort()` timeout to 0.5 seconds fixes the multiple `onAVStarted` calls.

I will do a couple of more tests today and remove the draft status very soon.

dagwieers:
A small visual fix could be that e.g. instead of showing in Up Next

> De Ideale Wereld - S2019 najaarE13 - 15/10 - Met Paula Sémer  2019-10-15

we drop the specific labeling that includes the date as well

> De Ideale Wereld - S2019 najaarE13 - Met Paula Sémer  2019-10-15

even better would be if in this case the season/episode looked a bit better, but not sure if we want to touch that. But at least the date could be appended in a better way:

> De Ideale Wereld - S2019 najaarE13 - Met Paula Sémer (2019-10-15)

dagwieers:
I got a weird error with one Terzake episode that always fails on Up Next.
This is episode **2019x200**: Brand en vandalisme tijdens protesten in Barcelona from 2019-10-18.

I always get:
```
2019-10-21 00:47:24.394 T:1170248560  NOTICE: [service.upnext] Monitor -> Calling autoplayback totaltime - playtime is 28.0849609375
2019-10-21 00:47:24.395 T:1170248560  NOTICE: [service.upnext] Api -> Got active player {u'jsonrpc': u'2.0', u'id': 1, u'result': [{u'playerid': 1, u'type': u'video', u'playertype': u'internal'}]}
2019-10-21 00:47:24.395 T:1170248560  NOTICE: [service.upnext] Api -> Getting details of now playing media
2019-10-21 00:47:24.429 T:1170248560  NOTICE: [service.upnext] Api -> Got details of now playing media {u'jsonrpc': u'2.0', u'id': 1, u'result': {u'item': {u'plotoutline': u'', u'episode': -1, u'season': -1, u'tvshowid': -1, u'label': u'638033162527', u'genre': [], u'playcount': 1, u'type': u'unknown', u'showtitle': u''}}}
2019-10-21 00:47:24.430 T:1170248560  NOTICE: [service.upnext] Api -> Got details of next up episode {u'jsonrpc': u'2.0', u'id': 1, u'error': {u'message': u'Invalid params.', u'code': -32602, u'data': {u'method': u'VideoLibrary.GetEpisodes', u'stack': {u'message': u'Value between 1 (inclusive) and 2147483647 (inclusive) expected but -1 received', u'type': u'integer', u'name': u'tvshowid'}}}}
2019-10-21 00:47:24.531 T:1170248560  NOTICE: [service.upnext] PlaybackManager -> Error: no episode could be found to play next...exiting
2019-10-21 00:47:24.532 T:1170248560  NOTICE: [service.upnext] Monitor -> Up Next style autoplay succeeded.
```
This is with Up Next v1.0.1 as well as the master branch.

dagwieers:
I also noticed that if you press **Watch Now** to go to the next episode before this episode has ended, the current episode will not be marked as watched (as it was prematurely ended). So I think we need to fix that as well.

dagwieers:
I also noticed that the thumbnail that is shown in Up Next is in fact that Program thumbnail, not the Episode thumbnail. I think that is a small improvement for our integration.

dagwieers:
I also added a new **jsonrpc()** function that I also implemented for Up Next.
This can be used in this PR as well.

resources/lib/service.py
```diff
@@ -20,6 +21,24 @@ def __init__(self):
...
+            import json
+            hexdata = json.loads(data)
+
+            if hexdata:
+                from binascii import unhexlify
+                data = json.loads(unhexlify(hexdata[0]))
+                _kodi.log('[Up Next]: %s, %s, %s' % (sender, method, to_unicode(data)), 'Verbose')
+
+                if 'plugin.video.vrt.nu_play_action' in method:
+                    executebuiltin('PlayerControl(Stop)')
```
dagwieers: I think it is preferred to use JSONRPC for this. https://kodi.wiki/view/JSON-RPC_API/v9#Player.Stop
I think you can replace this with:
```python
                jsonrpc(method='Player.Stop')
```

dagwieers: Hmm, it does require a `playerid`, which would require this:
```python
def get_playerid():
    ''' Get current playerid '''
    result = dict()
    while not result.get('result'):
        result = utils.jsonrpc(method='Player.GetActivePlayers')
    if not result.get('result'):
        return None
    return result.get('result')[0].get('playerid')
```

resources/lib/service.py
```diff
@@ -20,6 +21,24 @@ def __init__(self):
...
+                    executebuiltin('PlayMedia(plugin://plugin.video.vrt.nu/play/whatson/%s)' % data.get('whatson_id'))
```
dagwieers: And I guess this could become:
```python
                utils.jsonrpc(method='Addons.ExecuteAddon',  params=dict(addonid='plugin.video.vrt.nu', params=['/play/whatson/%(whatson_id)s/' % data]))
```

resources/lib/kodiwrapper.py
```diff
@@ -377,6 +381,20 @@ def get_global_setting(setting):
         json_result = xbmc.executeJSONRPC('{"jsonrpc": "2.0", "method": "Settings.GetSettingValue", "params": {"setting": "%s"}, "id": 1}' % setting)
         return json.loads(json_result).get('result', dict()).get('value')
 
+    def notify(self, sender, message, data):
+        ''' Send a notification to Kodi using JSON RPC '''
+        import json
```
dagwieers: This whole function could become:
```python
    def notify(self, sender, message, data):
        ''' Send a notification to Kodi using JSON RPC '''
        from binascii import hexlify
        import json
        result = self.jsonrpc(method='NotifyAll', params=dict(
            sender='%s.SIGNAL' % sender,
            message=message,
            data='["%s"]' % to_unicode(hexlify(json.dumps(data).encode())),
        ))
        if result.get('result') != 'OK':
            self.log_error('Failed to send notification: {error_message}', error_message=result.get('er
            return False
        self.log('Succesfully sent notification', 'Verbose')
        return True
```

resources/lib/vrtplayer.py
```diff
@@ -366,3 +382,14 @@ def push_position(self, info, video):
         # Push resumepoint to VRT NU
         self._resumepoints.update(uuid=uuid, title=title, url=url, watch_later=None, position=info.get('position'), total=info.get('total'))
         self._kodi.container_refresh()
+
+    def push_upnext(self, info):
+        ''' Push episode info to Up Next service add-on'''
+        if self._kodi.has_addon('service.upnext') and self._kodi.get_setting('useupnext', 'true') == 'true' and info.get('program'):
+            next_info = self._apihelper.get_upnext(info)
+            if next_info:
+                from binascii import hexlify
```
dagwieers: With that new proposed `notify()` function, the block would become a oneliner:
```python
self._kodi.notify(sender=self._kodi.addon_id(), message='upnext_data', data=next_info)
```

resources/lib/service.py
dagwieers: I tried various things, none of these would work:
```python
        _kodi.jsonrpc(method='Player.Stop', params=dict(playerid=_kodi.get_playerid()))
        _kodi.jsonrpc(method='GUI.ActivateWindow',  params=dict(
            window='videos',
            params=['plugin://plugin.video.vrt.nu/play/whatson/%(whatson_id)s/' % data],
        ))
```
```python
        _kodi.jsonrpc(method='Player.Stop', params=dict(playerid=_kodi.get_playerid()))
        _kodi.jsonrpc(method='Addons.ExecuteAddon',  params=dict(
            addonid='plugin.video.vrt.nu',
            params=['/play/whatson/%(whatson_id)s/' % data],
        ))
```
```python
        _kodi.jsonrpc(method='Player.Stop', params=dict(playerid=_kodi.get_playerid()))
        _kodi.jsonrpc(method='Player.Open', params=dict(item=dict(episodeid=data.get('whatson_id'))))
```

resources/lib/service.py
```diff
@@ -20,6 +21,24 @@ def __init__(self):
             if self.waitForAbort(10):
                 break
 
+    @staticmethod
+    def onNotification(sender, method, data):  # pylint: disable=invalid-name
+        ''' Handler for notifications '''
+        _kodi = KodiWrapper(None)
```
dagwieers: I rewrote this function as below:
```python
    @staticmethod
    def onNotification(sender, method, data):  # pylint: disable=invalid-name
        ''' Handler for notifications '''
        if not sender.startswith('upnextprovider'):
            return
        if not method.endswith('plugin.video.vrt.nu_play_action'):
            return

        import json
        hexdata = json.loads(data)

        if not hexdata:
            return
 
        from binascii import unhexlify
        data = json.loads(unhexlify(hexdata[0]))

        _kodi = KodiWrapper(None)
        _kodi.log('[Up Next]: %s, %s, %s' % (sender, method, to_unicode(data)), 'Verbose')
 
        executebuiltin('PlayerControl(Stop)')
        executebuiltin('PlayMedia(plugin://plugin.video.vrt.nu/play/whatson/%s)' % data.get('whatson_id'))
```

resources/lib/service.py
mediaminister: Done.

resources/lib/service.py
mediaminister: I tried
```python
        _kodi.jsonrpc(method='Player.Open', params=dict(item=dict(file='plugin://plugin.video.vrt.nu/play/whatson/%s' % data.get('whatson_id'))))
```
and it worked.

resources/lib/kodiwrapper.py
mediaminister: I moved the `'.SIGNAL'` and data encoding/hexlify parts back to vrtplayer because I want to be able to use `notify` with "normal" senders and json data. (for future use)
The Up Next add-on is using a specific notification format used in https://github.com/ruuk/script.module.addon.signals
I didn't want to add another dependency to the VRT NU add-on so I implemented the `notify` function without using `script.module.addon.signals`.

resources/lib/service.py
mediaminister: Added with working jsonrpc commands.

dagwieers:
Is there a reason to not merge this, even with the known issues ?

mediaminister:
I didn't merge it yet because I had to test the latest changes on Kodi 19. Testing is done now, so I'm going to merge this.

PR #527, opened 2019-10-14
Fix goto program in recent menu

mediaminister:


resources/lib/metadata.py
```diff
@@ -134,7 +134,7 @@ def get_context_menu(self, api_data, program, cache_file):
                     ))
 
         # Go to program context menu
-        if plugin.path.startswith(('/favorites/offline', '/favorites/recent', '/offline', 'recent', '/resumepoints/continue', '/resumepoints/watchlater', '/tvguide')):
+        if plugin.path.startswith(('/favorites/offline', '/favorites/recent', '/offline', '/recent', '/resumepoints/continue', '/resumepoints/watchlater', '/tvguide')):
```
dagwieers: Whoaw, thanks for noticing !

mediaminister: I urgently needed this feature while testing "Up Next".

PR #537, opened 2019-10-22
Only instantiate KodiWrapper when needed

dagwieers:
This is to avoid instantiating KodiWrapper for all events.


dagwieers:
I had this change also in my branch and it was the only change that isn't in master.

PR #538, opened 2019-10-22
Add HTTP error to dialog

dagwieers:
I got another HTTP Error 415: Unsupported Media Type, although we do not
report or log this error anywhere.

This change adds it to the dialog and logs it to the log. So that if it
gets reported (e.g. to VRT NU), hopefully the details will help.

PS I got this error with the HLS stream of episode 5 (season 4) of 24 hours in police custody. I reported this to VRT NU.

This PR includes:
- Put the HTTP Error in the dialog
- Log the HTTP Error to a log file
- Don't handle resumepoints if stream is None (and not played)

### DIFF
--- a/resources/lib/streamservice.py
+++ b/resources/lib/streamservice.py
@@ -250,13 +250,13 @@ class StreamService:
 
                 stream = StreamURLS(manifest_url, license_key=license_key, use_inputstream_adaptive=True)
             elif protocol == 'mpeg_dash':
+                self._kodi.log('Protocol: mpeg_dash', 'Verbose')
                 stream = StreamURLS(manifest_url, use_inputstream_adaptive=True)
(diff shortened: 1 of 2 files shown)


mediaminister:
Looks good to me.
To fix the root cause of this error, I should have a debug log or at least the url that returned http error 415

dagwieers:
It is a problem with VRT, they confirmed the issue which they are still figuring out.
(I assume it is in a third party product)

mediaminister:
I remember now, VRT uses Unified Origin and HTTP Error 415 is documented, VRT should indeed fix it: https://docs.unified-streaming.com/documentation/vod/index.html

dagwieers:
@mediaminister Maybe add this information to the open issue #307, this may help people finding our issue.

PR #543, opened 2019-10-23
WIP: Fix known cosmetic issues

dagwieers:
This PR includes:
- Disable menu caching for Main menu
- Disable menu caching for My favorites submenu
- Rename A-Z to 'All programs' (retained in icon and breadcrumbs)
- Change 'select' controls into 'enum' controls in settings

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -25,7 +25,7 @@ msgid "Browse only the programs you follow"
 msgstr ""
 
 msgctxt "#30012"
-msgid "A-Z"
+msgid "All programs"
 msgstr ""
 
 msgctxt "#30013"
@@ -480,6 +480,10 @@ msgstr ""
 
 msgctxt "#30433"
 msgid "Debug"
+msgstr ""
+
+msgctxt "#30440"
+msgid "A-Z"
 msgstr ""
 
 
--- a/resources/language/resource.language.nl_nl/strings.po
+++ b/resources/language/resource.language.nl_nl/strings.po
@@ -25,8 +25,8 @@ msgid "Browse only the programs you follow"
 msgstr "Bekijk alleen de tv-programma's die je volgt"
 
 msgctxt "#30012"
-msgid "A-Z"
-msgstr "A-Z"
+msgid "All programs"
+msgstr "Alle programma's"
 
 msgctxt "#30013"
 msgid "Alphabetically sorted list of programs"
@@ -481,6 +481,10 @@ msgstr "Meer"
 msgctxt "#30433"
 msgid "Debug"
 msgstr "Foutopsporing"
+
+msgctxt "#30440"
+msgid "A-Z"
+msgstr "A-Z"
 
 
 ### SETTINGS
--- a/resources/lib/addon.py
+++ b/resources/lib/addon.py
@@ -77,7 +77,7 @@ def favorites_menu():
 
 @plugin.route('/favorites/programs')
 def favorites_programs():
-    ''' The favorites A-Z listing '''
+    ''' The favorites 'My programs' listing '''
     from vrtplayer import VRTPlayer
     VRTPlayer(kodi).show_tvshow_menu(use_favorites=True)
 
@@ -146,7 +146,7 @@ def resumepoints_watchlater():
 @plugin.route('/programs/<program>')
 @plugin.route('/programs/<program>/<season>')
 def programs(program=None, season=None):
-    ''' The programs A-Z / seasons / episode listing '''
+    ''' The Programs / Seasons / Episodes listing '''
     from vrtplayer import VRTPlayer
     if program:
         VRTPlayer(kodi).show_episodes_menu(program=program, season=season)
--- a/resources/lib/apihelper.py
+++ b/resources/lib/apihelper.py
@@ -663,7 +663,10 @@ class ApiHelper:
                 # NOTE: Playcount is required to not have live streams as "Watched"
                 info_dict = dict(title=label, plot=plot, studio=channel.get('studio'), mediatype='video', playcount=0, duration=0)
                 stream_dict = dict(duration=0)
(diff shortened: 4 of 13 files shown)


dagwieers:
It would be nice if we could invalidate the menu caches e.g. when settings change. That would be more efficient than disabling menu caching for some of the directories that only change due to settings.

michaelarnauts:
Doesn't this already happen here https://github.com/pietje666/plugin.video.vrt.nu/blob/ff28b8393f8d0a21ba0297de4a1fe806b2363c3b/resources/lib/service.py#L53 ?

dagwieers:
@michaelarnauts Those are the HTTP caches that we manage ourselves. But Kodi also has a menu cache built-in (endOfDirectory), that we use for menus that do not change. But our main menu and favorites menu only changes when you modify settings (which is very infrequently) and currently we are forced to disable menu caching, while if we could invalidate the menu cache ourselves, those menus would be even faster when cached.

dagwieers:
In this PR I tried two things that did not work:
- Move the selection to a specific ListItem (for the TV Guide to default to Today)
- Add support for label2Mask (use %D when Unsorted) so we can keep the original title names

Both failed, even on Kodi v18.4. The latter test I need to remove.

mediaminister:
So, is it useful that I add commits to this branch? I'm fixing the "Universiteit van Vlaanderen" Up Next bug.

dagwieers:
@mediaminister I would fix bugs in separate PRs. It is going to be a lot easier if you have a working fix. And reference the PR numbers in the tracking issue so we can verify merged fixes work.

dagwieers:
Another test is to add `sys.exit()` to end the thread explicitly. However, I don't think that is actually needed, if stream_position ends the thread should end by itself. The question here is whether testing for `isPlaying()` every 500ms is really sufficient. If new videos are being played in quick succession, the threads could accumulate.

I think we should at least compare the file being played, or add a random number to the properties and verify if this has changed to know when to exit the thread..

Another option is to kill any pre-existing thread before starting a new one, so that at all times we know there is only one thread running (the latest one).

mediaminister:
> The question here is whether testing for isPlaying() every 500ms is really sufficient. If new videos are being played in quick succession, the threads could accumulate.

I think this is not possible, there is always a "stop" event that triggers the playerinfo thread to quit.
And when a new video is played, a new playerinfo object is created.

dagwieers:
So in order to avoid the additional `Container.Refresh` we need to make sure the internal resumepoints are updated when the directory gets refreshed by Kodi. I made the resumepoints update (online) async, and that increases our chances of this being the case, but it would be much better if we weren't reading the resume_points from disk almost every time we refresh e.g. Recent items in this case.

```
2019-10-26 19:18:16.914 T:1312813952  NOTICE: deleting video codec
2019-10-26 19:18:16.925 T:1937088528  NOTICE: VideoPlayer: finished waiting
2019-10-26 19:18:16.925 T:1937088528  NOTICE: CVideoPlayer::CloseFile()
2019-10-26 19:18:16.925 T:1937088528  NOTICE: VideoPlayer: waiting for threads to exit
2019-10-26 19:18:16.927 T:1937088528  NOTICE: VideoPlayer: finished waiting
2019-10-26 19:18:17.254 T:1937088528   ERROR: Control 55 in window 10025 has been asked to focus, but it can't
2019-10-26 19:18:17.633 T:1673438080  NOTICE: [plugin.video.vrt.nu] Got VRT NU Player info: {'position': 606.51104736328125, 'total': 780.00006103515625, 'event': u'playbackstopped'}
2019-10-26 19:18:17.634 T:1673438080  NOTICE: [plugin.video.vrt.nu] URL get: https://vrtnu-api.vrt.be/search?i=video&facets[videoId]=vid-5d08f592-c621-48dd-9942-0eb9da23f569&size=1
2019-10-26 19:18:18.010 T:1312813952  NOTICE: [plugin.video.vrt.nu] Access: plugin://plugin.video.vrt.nu/favorites/recent
2019-10-26 19:18:18.069 T:1312813952  NOTICE: [plugin.video.vrt.nu] Cache 'favorites.json' is fresh, expires in 4 minutes and 48 seconds.
2019-10-26 19:18:18.098 T:1312813952  NOTICE: [plugin.video.vrt.nu] Cache 'resume_points.json' is fresh, expires in 3 minutes and 19 seconds.
2019-10-26 19:18:18.109 T:1312813952  NOTICE: [plugin.video.vrt.nu] Cache 'my-recent-1.json' is fresh, expires in 48 minutes and 23 seconds.
2019-10-26 19:18:18.490 T:1673438080  NOTICE: [plugin.video.vrt.nu] Write cache 'resume_points.json'.
2019-10-26 19:18:18.855 T:1205302144  NOTICE: [plugin.video.vrt.nu] Got cached token '/storage/.kodi/userdata/addon_data/plugin.video.vrt.nu/tokens/user_XVRTToken.tkn'
2019-10-26 19:18:18.856 T:1205302144  NOTICE: [plugin.video.vrt.nu] URL post: https://video-user-data.vrt.be/resume_points/contentdamvrt20191010universiteitvanvlaanderen20192020a0009depotwp00160242
2019-10-26 19:18:18.856 T:1205302144  NOTICE: [plugin.video.vrt.nu] URL post data:: {"url": "/vrtnu/a-z/universiteit-van-vlaanderen/2019-2020/universiteit-van-vlaanderen-s2019-2020a9/", "position": 606.51104736328125, "total": 780.00006103515625, "watchLater": false}
```

As you can see after stopping playback at **19:18:16.914**, the Kodi refresh happens at **19:18:18.010** and access resumepoints at **19:18:18.098**, whereas we update resumepoints at **19:18:18.490** So we come 400ms too late for Kodi to pick up our information.

Lowering loop sleep from 500 to 50, does not impact this either. Kodi is always faster to refresh the container then we are with updating resumepoints.

The best option for us is to refresh the resumepoints internally (we need the assetPath or uuid for this) and avoid rereading the cache in this case. It would be nice if we could detect from the call if this is Kodi refreshing or not.

dagwieers:
Superseeded by #555

PR #547, opened 2019-10-24
Prepare for v2.3.0 release

dagwieers:

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,14 @@ leave a message at [our Facebook page](https://facebook.com/kodivrtnu/).
 ![](./resources/media/screenshot04.jpg)
 
 ## Releases
+### v2.3.0 (2019-11-15)
+- Add integration for Up Next add-on (@mediaminister)
+- Store resume points and watchlater at VRT NU (@dagwieers, @mediaminister)
+- Implement "Continue watching" (@mediaminister)
+- Implement "Watch later" (@dagwieers)
+- Fixes to "Go to program", "TV guide" and "Live TV" entries (@dagwieers)
+- Cosmetic changes to metadata and plot (@dagwieers)
+
 ### v2.2.3 (2019-10-01)
 - Integrate new inputstreamhelper (@dagwieers)
 - Support user paths with unicode characters (@mediaminister)
--- a/README.md
+++ b/README.md
@@ -50,13 +50,14 @@ leave a message at [our Facebook page](https://facebook.com/kodivrtnu/).
 ![](./resources/media/screenshot04.jpg)
 
 ## Releases
-### v2.3.0 (2019-11-15)
+### v2.3.0 (2019-12-31)
 - Add integration for Up Next add-on (@mediaminister)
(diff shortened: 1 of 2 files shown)


dagwieers:
IMO there is still a long way to go, but I would like to get more people to test this upcoming release. And we need a branch for this to get users to a zip file. So I will keep this branch in sync with the master branch.

dagwieers:
@mediaminister It appears that Kodi trims whitespaces from the start of the description/news fields, but doesn't do so for each consecutive line, only from the beginning and end of the string. That means that an indented payload in XML will become incorrectly indented in the Add-on information pane :-/

mediaminister:
Okay, I see.

mediaminister:
FYI, for "Up Next" to work with VRT NU add-on, you need at least Up Next version 1.0.2 which is not available in Kodi repo right now. The best option would be to make a new Up Next release first.

dagwieers:
@mediaminister What exactly do we need from Up Next v1.0.2 ?

mediaminister:
> What exactly do we need from Up Next v1.0.2 ?

I didn't investigated it yet, but while testing on Windows I found out that Up Next 1.0.1 doesn't play the next episode with Kodi 19 nightly.
I found no problems with 1.0.1 on Kodi 18.4

mediaminister:
I meanwhile investigated this and I found out Up Next needs a new release to support Python3 in Kodi 19. With older versions, there are different errors related to Python3.

I was mistaken that we only need 1.0.2, we really need a new 1.0.4 version.

dagwieers:
So, now that we improved PlayerInfo and the add-on is again faster (which was one of the reason I was looking into PlayerInfo before I realised today all video listings were slow since two weeks) we should think about releasing this.

No more testing new changes and improvements, but testing what we have and fixing issues :-)
(I'll put any ideas in the drawer until we have v2.3.0 out, I promise !)

dagwieers:
The only PR I would like to see in v2.3.0 is #591, but we depend on Up Next being released first.

dagwieers:
I fear this is going to become a Christmas release instead :-/

Hunting a strange unfollow issue now.

mediaminister:
> Hunting a strange unfollow issue now.

If you can provide more info, I'll take a look at it.

dagwieers:
You can follow a program just fine, but unfollowing a program simply doesn't work. It has no effect anymore.

mediaminister:
Issue with Unfollow is introduced in https://github.com/pietje666/plugin.video.vrt.nu/commit/cb68948028cc3ec5a9372ecc57d0741b9a61ac18

mediaminister:
Fixed unfollow issue in https://github.com/pietje666/plugin.video.vrt.nu/pull/617

dagwieers:
Up Next v1.0.7 is proposed for release, let's hope this gets released soon.
That would be one more item ticked off our list.

A list of things to be wary off is available from this tracker: #616

dagwieers:
Prepare for take-off.

PR #549, opened 2019-10-25
Fix Universiteit van Vlaanderen Up Next bug

mediaminister:


resources/lib/apihelper.py
```diff
@@ -269,7 +269,7 @@ def get_upnext(self, info):
         episodes = sorted(self.get_episodes(keywords=program), key=lambda k: (k.get('program'), k.get('seasonTitle'), k.get('episodeNumber')))
         upnext = dict()
         for episode in episodes:
-            if episode.get('program') == program and str(season) in episode.get('seasonTitle') and int(episode.get('episodeNumber')) == current_ep_no:
+            if episode.get('program') == program and episode.get('seasonTitle').startswith(str(season)) and int(episode.get('episodeNumber')) == current_ep_no:
```
dagwieers: I am surprised that this would cause the problem. ('2019' was not in seasonTitle '2018' ?)

Can we not just expect that `episode.get('seasonTitle') == str(season) ?

mediaminister: No, I get the `season` from `xbmc.Player().getVideoInfoTag().getSeason()` to save an extra lookup in VRT Search API and this Kodi function only supports integers...

This is a problem for `seasonTitles` with non-numeric characters:
universiteit-van-vlaanderen :: 2019-2020
europa-league-magazine :: 2019 NJ
de-ideale-wereld :: 2019 najaar
europa-league :: 2019 najaar
iedereen-beroemd :: 2019-2020
blokken :: 2019-2020
europa-league-preview-magazine :: 2019 NJ
extra-time :: 2019 najaar||2019-nj
ghost-rockers :: 1 compilatie
de-blacklist :: 2 compilatie
campus-12 :: 2 compilatie
a-touch-of-frost :: 1-6
ketnet-musical :: 4 compilatie

dagwieers: Ouch, maybe that needs to be added as a comment, because that is not obvious.

Is there no other way to add information to the Player(), or maybe add-on Window ?
We only need a thread-safe way to keep a current list of information of what is being played.

PS Inconsistent season names we ought to report, but non-integer season names we have to support :-/

mediaminister: I'll have to investigate this. I'll add a fix to this pull request when ready.

mediaminister: Fixed, it turned out that I didn't need to get the season and episode number from the player.

mediaminister:
Please test, Up Next should work flawlessly with all seasons and programs on VRT NU now.

dagwieers:
@mediaminister This I understand :-) Now that we store the path, is there no value in testing if the currently played file is the one we expect in the Thread? Or am I too paranoid and is there no issue at all?

I still need to test with threading debug information so I can track the different threads being started/stopped. For my own understanding, as this is not something we would like to see in Debug mode, I guess.

mediaminister:
When increasing the interval in the while loops to 500 ms, I added logging to the PlayerInfo class to test the timing and I found out the timing was okay, the previous thread was stopped before `onAVStarted` was called. I removed the logging before committing.

I'm okay with debug messages in the player events and outside the while loop or adding other tests.

dagwieers:
Works fine. More testing will be easier when it is merged.

Thanks !

PR #552, opened 2019-10-26
Implement Integration settings

### LINKED ISSUE #548: Remove optional requirements, create new Settings tab
### Describe the feature ###
It is probably better to remove all optional requirements from the add-on, and add specific entries in the settings to install, enable and configure these optional add-ons.

Not only is it more user-friendly, it offers a cleaner integration using a single user-interface.
Probably have a new Settings tab:

**Integration**
- Install YouTube add-on *(when not already installed)*
- Enable YouTube integration *(when installed)*
  - YouTube settings...
- Install Up Next add-on *(when not already installed)*
- Enable Up Next integration *(when installed)*
  - Up Next settings...
- Install PySocks  addon *(when not already installed)*

Since yesterday's problems at VRT NU, Iit could also be useful to have Twitter integration to show the most recent VRT NU twitter replies. It would have confirmed VRT NU issues almost immediately. How to do such integration is something yet to be discovered ;-) See #551

dagwieers:
This implements a new Integration tab in the settings.

To do this we had renumber a large part of the Add-on settings.

This fixes #548

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -1,12 +1,10 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon id="plugin.video.vrt.nu" name="VRT NU" version="2.2.3" provider-name="Martijn Moreel, dagwieers">
 <requires>
-  <import addon="plugin.video.youtube" version="6.5.1" optional="true"/>
   <import addon="resource.images.studios.white" version="0.0.22"/>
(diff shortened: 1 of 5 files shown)


dagwieers:
I expect more integrations to come in the future. Twitter is already being prepared.
Since we have implemented everything VRT NU exposes, the future is better integration...

PR #561, opened 2019-10-31
Use same margin for resumepoints and upnext

dagwieers:
This ensures the same margin is being used for both resumepoints and
upnext so that when moving to the next epside, the current episode is
considered watched.


dagwieers:
We should also update the resume position to the total time when signaled by upnextprovider.
(Both for resumepoints as well as `Files.SetFileDetails()`)

PR #562, opened 2019-10-31
Change the order of context menu items

dagwieers:
Re-evaluating the context menu, I think it makes more sense to put
'Watch later' first. The rationale is that for individual video's or
episodes this it more related to the context (episode) then
program-related context menu items.

For a program, there is no 'Watch later' item, so it starts with
'Follow program' and 'Go to program'.

And only then we have 'Refresh menu' which is not related to the context
at all.

So the order is: **Episode** > **Program** > **Menu**


dagwieers:
It would also make sense to skip the 'Follow program' for one-off videos, just as we do for 'Go to program'.

PR #563, opened 2019-11-03
Making settings instantaneous

dagwieers:
This PR doesn't store settings once on activation, but reevalutes
settings on use. I don't know what the performance impact is, but since
it is backed by C++ I would expect this to be efficient :-)


dagwieers:
With this change you can go to the settings enable/disable features and you will see the impact instantly when returning from the settings. The listing will e.g. update without favorites when you disable them, or without watching activity items when disabling resumepoints.

Just as one would expect. This was required after we enabled `reuselanguageinvoker`.

PR #565, opened 2019-11-03
Store the totaltime when playback has ended

### LINKED ISSUE #564: Forwarding to the end does not mark as Watched
### Describe the bug ###
Since resumepoints, forwarding to the end does not mark the video as watched. It appears that somehow forwarding to the end is not registered by the events we monitor.

The reason for this is that it registers the last stored resumepoint, and since we forwarded to the end, it never stored the end position. So we should be writing the end position when it has ended.

dagwieers:
This is to ensure forwarding to the end marks a video as watched.

This fixes #564


dagwieers:
With this fix I noticed we do no longer have to perform a Container.Refresh when the movie has ended. It is marked as watched immediately when returning !

This also makes me think the indicate that something has been partly watch (which does require a container refresh) is broken in Kodi itself.

PR #570, opened 2019-11-07
Use Container.Update instead of Container.Refresh

mediaminister:
This should preserve the path history, so you can return to the previous container when navigating after playing a video.


dagwieers:
That was quick :smile:

mediaminister:
FYI, it's explained in the docs: https://codedocs.xyz/AlwinEsch/kodi/page__list_of_built_in_functions.html

Issue #560, opened 2019-10-30
Changes to settings do not take effect immediately


dagwieers:
### Describe the bug ###
Most of the changes to the settings require you to restart the add-on (which means first starting another add-on, then return to VRT NU). This is not very user-friendly (i.e. most people would think restarting the add-on would do the trick).

So it's safe to say that people would consider this a bug (changing settings doesn't have any effect). And especially for logging this is problematic, if you enable debug-logging in Kodi, our add-on wouldn't act on it.

We need to address this before releasing v2.3.0.

### To Reproduce ###
1. Change logging settings
2. Continue using the add-on
3. Or restart the add-on
4. Logging changes do not have any effect

### Expected behavior ###
We need to rethink this. Either reevaluate config-options on use (slower) or reevaluate settings when changes are detected.

### Additional context ###
- Operating system: LibreELEC 9.2 Beta2 (9.1.502)
- Kodi version: 18.4
- Addon version: 2.3.0 (devel branch)
- Using a VPN: no
- Country you are using the addon from: BE

mediaminister:
This is not fixed yet.  When enabling and disabling `useresumepoints` setting I get wrong values which causes problems with menu refreshing and the Kodi watched status.

I think the pluginsource entrypoint should be restructured according to Kodi's recommendations:
> The `<reuselanguageinvoker>` element is a feature introduced with Kodi 18.0 that changes the way the python invoker works in Kodi - trying to reuse the invoker instances as much as possible. As a result, the addon performance is greatly improved. **However, note that for the element to work some changes may be required in your addon. Namely, since the invoker is reused, make sure sys.argv is always passed to your entrypoint and propagated throughout your codebase. Do not store it as a class variable.**

mediaminister:
@dagwieers I found out that removing the global `kodi` variable in `addon.py` fixes this problem, but I'm not sure I have the right fix, the following looks rather strange to me:
```
VRTPlayer(KodiWrapper(globals())).show_main_menu()
```

Any suggestions?

dagwieers:
So one solution might be that:
- `get_setting()` reads from the Window property cache first (and uses ADDON second)
  - (and also updates the Window property cache if it access ADDON settings)
- service.py populates the Window property cache when `onSettingsChanged`
  - iterating over all known settings

The drawback here is that we add quite some complexity for something that really ought to be something simple.

dagwieers:
[...]

PR #574, opened 2019-11-13
Declare KodiWrapper locally when using reuselanguageinvoker

mediaminister:
This makes sure all `KodiWrapper` functions are executed instantaneous and return fresh values when using `reuselanguageinvoker`

This fixes https://github.com/pietje666/plugin.video.vrt.nu/issues/560

This makes realtime toggling the "watching activity" setting possible. (Should also work for all other settings)

### DIFF
--- a/resources/lib/addon.py
+++ b/resources/lib/addon.py
@@ -15,13 +15,14 @@ from statichelper import from_unicode, to_unicode
 
 # pylint: disable=invalid-name
 plugin = Plugin()
-kodi = KodiWrapper(globals())
+glob = globals()
 
 
 @plugin.route('/')
 def main_menu():
     ''' The VRT NU plugin main menu '''
     from vrtplayer import VRTPlayer
+    kodi = KodiWrapper(glob)
     VRTPlayer(kodi).show_main_menu()
 
 
@@ -29,6 +30,7 @@ def main_menu():
 @plugin.route('/cache/delete/<cache_file>')
 def delete_cache(cache_file='*.json'):
     ''' The API interface to delete caches '''
+    kodi = KodiWrapper(glob)
     kodi.refresh_caches(cache_file=cache_file)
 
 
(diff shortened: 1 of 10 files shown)


resources/lib/addon.py
```diff
@@ -15,131 +15,130 @@
...
     from vrtplayer import VRTPlayer
-    VRTPlayer(kodi).show_continue_menu(page=1)
+    VRTPlayer(KodiWrapper(globals())).show_continue_menu(page=1)
 
 
 @plugin.route('/resumepoints/refresh')
 def resumepoints_refresh():
     ''' The API interface to refresh the resumepoints cache '''
     from resumepoints import ResumePoints
-    ResumePoints(kodi).refresh(ttl=0)
```
michaelarnauts: You will only need to do this once per function.

michaelarnauts:
I've had a similar issue in VTM GO, since the Kodi wrapper is actually only initialised once, the old settings that are read in the first __init__ call are being used. Therefore, you need to read the setting every time instead of using the cached value.

michaelarnauts:
Hmm, just looked at it, and this doesn't seem to be the case. Ignore my comment :)

resources/lib/addon.py
mediaminister: Indeed, I did a quick find and replace. Fixed it.

mediaminister:
Thanks for reporting. If VTM GO add-on is not affected, there might be a cleaner fix for this problem.

dagwieers:
If we would instantiate in addon_entry.py and then pass it on, would that not help?

Personally I would use a global variable rather than a KodiWrapper object that we have to pass on.
The object is always needed anyway, so I don't see a good reason not to define it as global.

And then kodiwrapper.py would just hold a list of functions (something that would be more portable than what we currently have).

dagwieers:
@mediaminister The actual fix is in #575, the problem is that the settings seems to be read only when Addon() is instantiated, which means you don't see any updates to the settings for a reused Addon() instance.

PR #575, opened 2019-11-15
Replace kodiwrapper with kodiutils

### LINKED ISSUE #560: Changes to settings do not take effect immediately

dagwieers:
This is an attempt to simplify the addon.

This PR includes:
- Removing the KodiWrapper class in favor of discrete functions (kodiutils)
- Rename some kodiutils functions to shorter variants
- Selective imports from kodiutils and json
- Make more use of `has_addon()`
- Remove unused code (resumepoints/favorites invalidate_caches)
- Instantiate Addon() every time, don't cache it (this fixes #560)

This fixes #560


dagwieers:
From **plugin.video.mediathekview**:
```python
    def reload_settings(self):
        """ Reload settings and reconfigure service behaviour """
        # self.info("===== RELOAD SETTINGS =====")
        # TODO: support online reconfiguration
        #       currently there is a bug in Kodi: this event is only
        #       triggered if the reconfiguration happen inside the
        #       addon (via setSetting). If teh user changes something
        #       via the settings page, NOTHING WILL HAPPEN!
        pass
```

mediaminister:
I just did a quick test and I can confirm this fixes #560.
You can merge this already, I will test the master branch later this weekend.

dagwieers:
Do you agree with the removal of KodiWrapper as well ?

mediaminister:
Yes, I totally agree with removing KodiWrapper. I think simplifying the addon structure is very much needed.

PR #579, opened 2019-11-18
Improve stubs

michaelarnauts:
I've made a few changes for VTM Go, that might also be useful here:

* Allows tests to run locally with only Python 3.7 installed (recent Ubuntu versions)
* Allow specifying a query param in the `test/run` command
* Add `Dialog.contextmenu` stub
* Add `DialogProgressBG` stub
* Add `ListItem`'s `setIsFolder` and `setProperties` functions (Kodi 18)
* Add missing `SORT_*` constants


dagwieers:
@michaelarnauts At some point I would like to split off our own stubs, but it would need a high level of customizations for some environments and I don't know how to offer that. E.g. for Up Next we would need to be able to test-drive scenarios where we fake playing something, and then have some events. So I think we first need to figure out how to do this in  way that it is generally useful and applicable (possibly with extensions).

I am also very much interested in improving our tests to query internal objects so when we run a specific URL, we can query the stubs if it was successful, whether the listing was not-empty, etc. Currently our integration tests are mostly visual tests, and if things would go wrong we may not know, but if we could test for correctness in the integration tests that would be massive.

I also would like to release the **kodiutils** library as an abstraction layer to facilitate writing add-ons. But also here it needs to be easily customizable so that our users can modify and ship a custom routine until their patches have been released in the Kodi repository. (Otherwise new functionality would become a blocking factor really quickly). As an independent module it would also be easier to test everything in isolation, and unused functionality would not mean lower coverage ;-)

Main focus would be on performance and not requiring external python libraries.

Makefile
```diff
@@ -1,4 +1,4 @@
-ENVS := flake8,py27,py36
+ENVS := flake8,py27,py36,py37
```
dagwieers: The reason why I haven't added this is that on my RHEL7 laptop I don't have a working Python 3.7 environment, and adding this makes Tox fail. Maybe there is a way to no fail on `InterpreterNotFound`.

test/run.py
```diff
@@ -28,5 +28,11 @@
 if not sys.argv[1].startswith('plugin://'):
     sys.argv[1] = 'plugin://plugin.video.vrt.nu' + sys.argv[1]
 
-print('** Running URI: %s' % sys.argv[1])
-addon.run([sys.argv[1], 0, ''])
+# Split path and args
+try:
+    path, args = sys.argv[1].split('?', 1)
```
dagwieers: So I really like exception handling in general, but try to avoid it in those cases where it is easy to avoid.
And the main reason is that exception handling has a performance impact (not that it would matter here)

This would do:
```python
path, _, args = sys.argv[1].partition('?')
```

test/xbmcgui.py
```diff
@@ -191,6 +232,11 @@ def setInfo(type, infoLabels):  # pylint: disable=redefined-builtin
         ''' A stub implementation for the xbmcgui ListItem class setInfo() method '''
         return
 
+    @staticmethod
+    def setIsFolder(isFolder):
```
dagwieers: So this one does not exist in Kodi Krypton. And for VRT NU we are testing also for Krypton.
Maybe we would need to add e.g. a decorator so functions or methods that are not supported would trigger and error.
And at the same time we need a mechanism to indicate the lowest version of Kodi that is supported.

```python
@leia
@staticmethod
def setIsFolder(isFolder):
```

test/run.py
dagwieers: I'll modify and merge.

test/xbmcgui.py
```diff
@@ -205,6 +251,11 @@ def setProperty(key, value):
         ''' A stub implementation for the xbmcgui ListItem class setProperty() method '''
         return
 
+    @staticmethod
+    def setProperties(dictionary):
```
dagwieers: Same here, does not exist in Kodi Krypton.

PR #583, opened 2019-11-20
Use video_id for Up Next because whatson_id is not unique

mediaminister:
While testing I found out `whatson_id` is not unique. Some episodes have multiple versions on VRT NU, versions with or without audio description or Flemish Sign Language.
It turns out these versions have the same `whatson_id`. 
For instance: Dag Sinterklaas: https://vrtnu-api.vrt.be/search?i=video&facets[whatsonId]=762744552527

This causes Up Next to play a next episode with audio description while previous episodes were played without audio description.

### DIFF
--- a/resources/lib/apihelper.py
+++ b/resources/lib/apihelper.py
@@ -333,7 +333,7 @@ class ApiHelper:
             )
 
             play_info = dict(
-                whatson_id=next_ep.get('whatsonId'),
+                video_id=next_ep.get('videoId'),
             )
 
             next_info = dict(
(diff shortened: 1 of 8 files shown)


dagwieers:
That makes whatsonid not very useful. Can we make a note about that on the VRT NU API wiki page?
I guess we need to update that page wrt. to other information as well at some point.

resources/lib/apihelper.py
```diff
@@ -353,10 +353,13 @@ def get_upnext(self, info):
...
         return None
 
-    def get_single_episode(self, whatson_id):
-        ''' Get single episode by whatsonId '''
+    def get_single_episode(self, ep_id, upnext=False):
+        ''' Get single episode by whatsonId or videoId '''
         video = None
-        api_data = self.get_episodes(whatson_id=whatson_id, variety='single')
+        if upnext:
+            api_data = self.get_episodes(video_id=ep_id, variety='single')
```
dagwieers: I am not so happy with this interface. The upnext boolean is actually a switch to decide whether it is a video_id or a whatson_id. I much rather have a more explicit interface:
```python
def get_single_episode(self, video_id=None, whatson_id=None):
```

*Or possibly two entry points instead of a shared one.*

mediaminister: Okay, changed.

mediaminister:
> That makes whatsonid not very useful. Can we make a note about that on the VRT NU API wiki page?

I already added the following info to the wiki:
- **whatsonId**
  - not a unique identifier, correspondends to a timeslot in a broadcast schedule: https://en.wikipedia.org/wiki/WHATS%27On_(Software)
  - episodes with multiple versions (for audio description or Flemish Sign Language) have the same whatsonId

PR #586, opened 2019-11-23
This fixes a known issue on Python 3.5

dagwieers:
Apparently Python 3.5 may be common using Debian Stretch (OSMC).

### DIFF
--- a/resources/lib/service.py
+++ b/resources/lib/service.py
@@ -66,7 +66,12 @@ class VrtMonitor(Monitor):
             return
 
         from binascii import unhexlify
-        data = loads(unhexlify(hexdata[0]))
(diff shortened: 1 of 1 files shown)


michaelarnauts:
Hmm, I thought Kodi used an embedded Python interpretor (2.7 for Kodi 18 and 3.6 for Kodi 19).

Does it makes sense then to test for Python 3.5 stuff? Or maybe even for Python 3.8 things?

dagwieers:
Yup, I think we need to test for Python 3.5.
We already test for Python 3.8.

mediaminister:
> Hmm, I thought Kodi used an embedded Python interpretor
This may be the case for Windows, but not for Linux.

PR #597, opened 2019-11-28
Improve playerinfo

mediaminister:
This PR includes:
- Add window properties functions
- Store Container.URL in Window property
- Simplify Service class and improve PlayerInfo class
- Rely on Kodi watch status when possible and reduce container updates

### DIFF
--- a/resources/lib/kodiutils.py
+++ b/resources/lib/kodiutils.py
@@ -139,6 +139,7 @@ def show_listing(list_items, category=None, sort='unsorted', ascending=True, con
     from addon import plugin
     from xbmcgui import ListItem
 
+    set_property('container.url', 'plugin://' + addon_id() + plugin.path)
     xbmcplugin.setPluginFanart(handle=plugin.handle, image=from_unicode(addon_fanart()))
 
     usemenucaching = get_setting('usemenucaching', 'true') == 'true'
@@ -407,6 +408,24 @@ def get_global_setting(key):
     ''' Get a Kodi setting '''
     result = jsonrpc(method='Settings.GetSettingValue', params=dict(setting=key))
     return result.get('result', {}).get('value')
+
+
+def get_property(key, window_id=10000):
+    ''' Get a Window property '''
+    from xbmcgui import Window
+    return to_unicode(Window(window_id).getProperty(key))
+
+
+def set_property(key, value, window_id=10000):
+    ''' Set a Window property '''
+    from xbmcgui import Window
+    return Window(window_id).setProperty(key, from_unicode(str(value)))
+
+
+def clear_property(key, window_id=10000):
+    ''' Clear a Window property '''
+    from xbmcgui import Window
+    return Window(window_id).clearProperty(key)
 
 
 def notify(sender, message, data):
--- a/resources/lib/metadata.py
+++ b/resources/lib/metadata.py
@@ -192,13 +192,11 @@ class Metadata:
 
                 position = self._resumepoints.get_position(asset_id)
                 total = self._resumepoints.get_total(asset_id)
+                # Master over Kodi watch status
                 if position and total and SECONDS_MARGIN < position < total - SECONDS_MARGIN:
                     properties['resumetime'] = position
+                    properties['totaltime'] = total
                     log(2, '[Metadata] manual resumetime set to %d' % position)
-
-            duration = self.get_duration(api_data)
-            if duration:
-                properties['totaltime'] = duration
 
             episode = self.get_episode(api_data)
             season = self.get_season(api_data)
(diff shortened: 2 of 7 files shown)


resources/lib/kodiutils.py
```diff
@@ -409,6 +410,24 @@ def get_global_setting(key):
...
+def get_property(key, window_id=10000):
+    ''' Get a Window property '''
+    from xbmcgui import Window
+    return Window(window_id).getProperty(key)
+
+
+def set_property(key, value, window_id=10000):
+    ''' Set a Window property '''
+    from xbmcgui import Window
+    return Window(window_id).setProperty(key, value)
```
dagwieers: For Up Next I already improved upon this:
```suggestion
    return Window(window_id).setProperty(key, from_unicode(str(value)))
```

resources/lib/kodiutils.py
```diff
@@ -409,6 +410,24 @@ def get_global_setting(key):
     return result.get('result', {}).get('value')
 
 
+def get_property(key, window_id=10000):
+    ''' Get a Window property '''
+    from xbmcgui import Window
+    return Window(window_id).getProperty(key)
```
dagwieers: For Up Next I already improved upon this:
```suggestion
    return to_unicode(Window(window_id).getProperty(key))
```

resources/lib/metadata.py
```diff
@@ -173,6 +173,8 @@ def get_playcount(self, api_data):
                 total = self._resumepoints.get_total(asset_id)
                 if position and total and position > total - SECONDS_MARGIN:
                     playcount = 1
+                elif position and total and position < SECONDS_MARGIN:
```
dagwieers: This doesn't seem right. We don't know the playcount is 0 in this case, and it is better to not set it and fall back on what Kodi already knows. (unknown will be zero anyway)

resources/lib/metadata.py
```diff
@@ -192,13 +194,15 @@ def get_properties(self, api_data):
...
                     properties['resumetime'] = position
+                    properties['totaltime'] = total
                     log(2, '[Metadata] manual resumetime set to %d' % position)
-
-            duration = self.get_duration(api_data)
-            if duration:
-                properties['totaltime'] = duration
+                else:
+                    # Slave to Kodi watch status
+                    properties['resumetime'] = -1
```
dagwieers: What is the benefit of doing this over not setting it ?
Also, I wasn't sure if setting the totaltime here has any value.

resources/lib/playerinfo.py
```diff
@@ -4,83 +4,149 @@
...
+        self.push_position(position=self.last_pos, total=self.total)
+
+    def onPlayBackResumed(self):  # pylint: disable=invalid-name
+        ''' Called when user resumes a paused file or a next playlist item is started '''
+        suffix = 'after pausing' if self.paused else 'after playlist change'
+        log(3, '[PlayerInfo] %d onPlayBackResumed %s' % (self.thread_id, suffix))
+        if not self.paused:
+            self.push_position(position=self.last_pos, total=self.total)
+        self.paused = False
+
```
dagwieers: I noticed you removed the onPlaybackSeek(), which is important to ensure we have updated `self.last_pos`.
Especially for doing things like, skipping to the end, or skipping to the start and stopping.
(Which is currently the only way to mark-as-watched, or reset position.

resources/lib/metadata.py
mediaminister: This is needed to overwrite the Kodi watch status so it matches VRT NU resumepoint positions.

dagwieers: But if you watch it first completely once, and then watch it again (even in the first 30 seconds), playcount would be 0. I don't think we need to explicitly reset playcount in the listitem in this case.

resources/lib/metadata.py
mediaminister: You're right I think, I'll remove this.

resources/lib/playerinfo.py
mediaminister: I'll add this just to be sure, but I think `onPlayBackEnded` also updates the last position when seeking to the end of the video.

dagwieers: Indeed, when the playback ended, but not when it was stopped.
To be clear, I don't want to update the position on seek, just set `self.last_pos` for in case it is used.

resources/lib/metadata.py
mediaminister: Indeed, that was the purpose. But I have removed this because I'm not sure what is best for the user keeping in mind that VRT NU resumepoints breaks the watched/unwatched context menu.

PR #601, opened 2019-11-29
Another Python35 json fix

mediaminister:


dagwieers:
So I would like to get rid of Python 3.5-specific changes when we stop supporting Python 3.5.
Not sure what the best way is to get reminded of this.

mediaminister:
This change is not really Python 3.5 specific, I just pass unicode text into `json.loads` and this approach happens to work for all Python 3.x versions.

A wrapper function that takes an url and outputs unicode json can simplify maintenance in this specific case. Using the `requests` module is also an option. ;-)

dagwieers:
urllib3 is a lot faster, but still not nearly as efficient as it used to be. Regardless of the ease it brings. However, there's a lot to say to make a separate module for all HTTP-related functionality (incl. cookie-handling and JSON parsing). Or maybe use a more efficiënt Requests alternative?

For #385 this would be very convenient too :-)

dagwieers:
> This change is not really Python 3.5 specific, I just pass unicode text into json.loads and this approach happens to work for all Python 3.x versions.

My only concern here was that rather than reading from a handle, this passes the content around, so not as efficiënt as before. I guess when we get rid of Python 2 *and* Python 3.5 we can clean this up.

PR #602, opened 2019-11-29
Only process PlayerInfo events from VRT NU

mediaminister:
Currently, PlayerInfo processes every play-action in Kodi, this PR ensures that PlayerInfo only processes `plugin://plugin.video.vrt.nu/` paths.

### DIFF
--- a/resources/lib/playerinfo.py
+++ b/resources/lib/playerinfo.py
@@ -19,6 +19,7 @@ class PlayerInfo(Player):
     def __init__(self):
         ''' PlayerInfo initialisation '''
         self.path = None
+        self.listen = False
         self.resumepoints = ResumePoints()
         self.apihelper = ApiHelper(Favorites(), self.resumepoints)
         self.paused = False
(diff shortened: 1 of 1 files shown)


dagwieers:
Good, this was causing exceptions on playback from other add-ons or Kodi.

I was looking at this before but couldn't find an acceptable implementation. I considered your implementation, but somehow I felt that PlayerInfo did not belong in the add-on service but would be better started when we initiate the playback ourselves.

It would mean we start our own PlayerInfo (Player) thread when we start playback. You probably know/remember why that was not an option. I would have to test this it to find that brick wall again :laughing:

dagwieers:
That said, I don't mind merging this as this is an improvement to what we have by any standard.

mediaminister:
Stopping and starting `PlayerInfo` was too complicated in combination with the implementation of the "Up Next" add-on. I got multiple concurrent PlayerInfo threads because I couldn't stop the previous thread fast enough. I believe this approach is still possible but needs a lot of work to get it right.

Another approach is completely relying on `onNotification` in `service.py`, so we know the sender add-on already when receiving the xbmc.Player event.

PR #607, opened 2019-11-30
Re-initialize addon when settings change

mediaminister:
This fixes https://github.com/pietje666/plugin.video.vrt.nu/issues/560

### DIFF
--- a/resources/lib/kodiutils.py
+++ b/resources/lib/kodiutils.py
@@ -113,6 +113,12 @@ def addon_profile():
     return to_unicode(xbmc.translatePath(ADDON.getAddonInfo('profile')))
 
 
+def reinitialize_addon():
+    ''' Re-initialize addon '''
+    global ADDON  # pylint: disable=global-statement
+    ADDON = Addon()
+
+
 def url_for(name, *args, **kwargs):
     ''' Wrapper for routing.url_for() to lookup by name '''
     import addon
@@ -377,9 +383,6 @@ def localize_from_data(name, data):
 
 def get_setting(key, default=None):
     ''' Get an add-on setting '''
-    # global ADDON  # pylint: disable=global-statement
-    # from xbmcaddon import Addon # pylint: disable=redefined-outer-name,reimported
-    # ADDON = Addon()
     try:
         value = to_unicode(ADDON.getSetting(key))
     except RuntimeError:  # Occurs when the add-on is disabled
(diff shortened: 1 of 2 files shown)


dagwieers:
Does this work?

Because both are running in two different 'languageinvokers' (python instances) I would assume. And each has a different process id, or am I wrong?

mediaminister:
I tested and it works. Approved by Team Kodi: https://forum.kodi.tv/showthread.php?tid=290353&pid=2425543#pid2425543

dagwieers:
It does not seem to work for me though.
If I open you addon in the Main menu, then go to settings and disable "Enable My favorites", then I would expect the Main menu to not show "My favorites", but it does.
Even restarting the add-on does not work, you have to load another add-on and then go back to VRT NU to ensure the new settings are updated.

dagwieers:
If this was possible, it would mean the service, and the addon would be in the same process space using the same module cache. And then the service could update addon objects, and vice versa.

But that is not possible, the only way service and addon can communicate is through Kodi itself (i.e. Player object, or Window properties, or notifications). Otherwise we wouldn't have had so many issues with Playerinfo and getting information across.

mediaminister:
I noticed strange behaviour. The first and second time it doesn't seem to work, but from the third time it works.

dagwieers:
I have had my share of strange behavior putting me on the wrong foot many times with Kodi.
(Combinations of reuselanguageinvoker, old service process and brainfarts)

But at least the PID of the addon and the service appear to be the same:
```
2019-11-30 01:58:17.738 T:1305473920  NOTICE: [plugin.video.vrt.nu] Service started (PID=648)
2019-11-30 01:58:20.570 T:1247798144  NOTICE: [plugin.video.vrt.nu] Access: plugin://plugin.video.vrt.nu/
2019-11-30 01:58:20.570 T:1247798144  NOTICE: [plugin.video.vrt.nu] PID = 648
```

So one assumption out the window already ;-) But restarting the add-on does not change the PID.
This PID however is the PID of the main **kodi.bin**, so I think this is because it is multi-threaded and the service and addon live in a different thread.

dagwieers:
> I noticed strange behaviour. The first and second time it doesn't seem to work, but from the third time it works.

It never works for me, can you give me a walkthrough how to see it work?

dagwieers:
I am open to improve #608 with something that doesn't update ADDON on every invocation. But at least this is slightly better than disabling `reuselanguageinvoker`.

All-in-all, I think this is something that Kodi should fix. Not us.

mediaminister:
> It never works for me, can you give me a walkthrough how to see it work?

It worked when you changes "Enable My favorites" three times in a row. (Three times opening and closing addon settings.)

But nevermind, it's fixed now in master branch.

dagwieers:
I did try that. It could be useful to understand such a mechanism, because it could be useful for other uses as well.

PR #608, opened 2019-11-30
Renew ADDON on every invocation

### LINKED ISSUE #560: Changes to settings do not take effect immediately

dagwieers:
This is the easiest solution for ensuring we have up-to-date settings.
But it incurs a small performance hit on every invocation.

This fixes #560

### DIFF
--- a/resources/lib/addon.py
+++ b/resources/lib/addon.py
@@ -13,8 +13,7 @@ except ImportError:  # Python 2
 from kodiutils import localize, log_access, notification, refresh_caches
 from statichelper import from_unicode, to_unicode
 
-# pylint: disable=invalid-name
-plugin = Plugin()
+plugin = Plugin()  # pylint: disable=invalid-name
 
 
 @plugin.route('/')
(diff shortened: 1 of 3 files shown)


dagwieers:
So after 10 different iterations I found this to work.
No idea why everything before didn't, but at least this is quite simple (unlike some attempts before this one).

PR #610, opened 2019-11-30
Seamlessly play Up Next episodes

mediaminister:
Sometimes an "Up Next" episode fails to play because the previous episode caused an `onPlayBackEnded` event while the next episode is still setting up. I experienced this while watching Peulengaleis episodes.

This pull request fixes this problem.

### DIFF
--- a/resources/lib/playerinfo.py
+++ b/resources/lib/playerinfo.py
@@ -4,7 +4,7 @@
 
 from __future__ import absolute_import, division, unicode_literals
 from threading import Event, Thread
-from xbmc import getInfoLabel, Player
+from xbmc import getInfoLabel, Player, PlayList
 from apihelper import ApiHelper
 from favorites import Favorites
 from resumepoints import ResumePoints
@@ -22,12 +22,14 @@ class PlayerInfo(Player):
         self.last_pos = None
         self.listen = False
         self.paused = False
-        self.total = 900
+        self.total = None
+        self.positionthread = None
         self.quit = Event()
 
         self.asset_id = None
         self.path = None
         self.title = None
+        self.ep_id = None
         self.url = None
         self.whatson_id = None
         from random import randint
@@ -49,20 +51,19 @@ class PlayerInfo(Player):
         # Get asset_id, title and url from api
         ep_id = play_url_to_id(self.path)
 
-        if ep_id.get('video_id'):
-            episode = self.apihelper.get_episodes(video_id=ep_id.get('video_id'), variety='single')[0]
-        elif ep_id.get('whatson_id'):
-            episode = self.apihelper.get_episodes(whatson_id=ep_id.get('whatson_id'), variety='single')[0]
-        elif ep_id.get('video_url'):
-            episode = self.apihelper.get_episodes(video_url=ep_id.get('video_url'), variety='single')[0]
+        # Get episode data
+        episode = self.apihelper.get_single_episode_data(video_id=ep_id.get('video_id'), whatson_id=ep_id.get('whatson_id'), video_url=ep_id.get('video_url'))
 
         self.asset_id = self.resumepoints.assetpath_to_id(episode.get('assetPath'))
         self.title = episode.get('program')
         self.url = url_to_episode(episode.get('url', ''))
+        self.ep_id = 'S%sE%s' % (episode.get('seasonTitle'), episode.get('episodeNumber'))
         self.whatson_id = episode.get('whatsonId') if episode.get('whatsonId') else None
-
         self.update_position()
-        self.push_position(position=self.last_pos, total=self.total, event='onPlayBackStarted')
+        self.update_total()
+
+        # Reset VRT NU resumepoint
+        self.push_position()
 
     def onAVStarted(self):  # pylint: disable=invalid-name
         ''' Called when Kodi has a video or audiostream '''
@@ -72,8 +73,13 @@ class PlayerInfo(Player):
         self.quit.clear()
         self.update_position()
         self.update_total()
+        self.push_upnext()
 
-        Thread(target=self.stream_position, name='StreamPosition').start()
+        # StreamPosition thread keeps running when watching multiple episode with "Up Next"
+        # only start StreamPosition thread when it doesn't exist yet.
+        if not self.positionthread:
+            self.positionthread = Thread(target=self.stream_position, name='StreamPosition')
+            self.positionthread.start()
 
     def onAVChange(self):  # pylint: disable=invalid-name
         ''' Called when Kodi has a video, audio or subtitle stream. Also happens when the stream changes. '''
@@ -96,7 +102,7 @@ class PlayerInfo(Player):
             return
         log(3, '[PlayerInfo %d] Event onPlayBackPaused' % self.thread_id)
         self.update_position()
-        self.push_position(position=self.last_pos, total=self.total, event='onPlayBackPaused')
+        self.push_position(position=self.last_pos, total=self.total)
         self.paused = True
 
     def onPlayBackResumed(self):  # pylint: disable=invalid-name
@@ -106,8 +112,7 @@ class PlayerInfo(Player):
         suffix = 'after pausing' if self.paused else 'after playlist change'
         log(3, '[PlayerInfo %d] Event onPlayBackResumed %s' % (self.thread_id, suffix))
         if not self.paused:
-            self.update_position()
-            self.push_position(position=self.last_pos, total=self.total, event='onPlayBackResumed')
+            self.push_position(position=self.last_pos, total=self.total)
         self.paused = False
 
     def onPlayBackEnded(self):  # pylint: disable=invalid-name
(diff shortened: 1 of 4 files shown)


dagwieers:
It is unfortunate if this is the best way to fix it, because it adds even more complexity by using Playlists IMO. For me it is impossible to understand the impact as I have no experience/insight in how Playlist works.

mediaminister:
Adding the next episode to the `PlayList` prevents  `onPlayBackEnded` getting triggered when the previous episode reaches the end of the file. And this also prevents that Kodi returns back to the menu.

I think this is the best way to seamlessly play consecutive episodes.

mediaminister:
> It is unfortunate if this is the best way to fix it, because it adds even more complexity by using Playlists IMO.

Unfortunately, I didn't find another way to prevent Kodi Player from closing and returning to the menu after every episode when binge watching complete series without user interaction.

I'm convinced this branch improves the user experience a lot when bingewatching VRT NU series.

This pull request is ready to be reviewed.

dagwieers:
I am not going to argue as I simply trust your opinion more than my lack of understanding :wink:

resources/lib/playerinfo.py
```diff
@@ -22,12 +22,14 @@ def __init__(self):
         self.last_pos = None
         self.listen = False
         self.paused = False
-        self.total = 900
+        self.total = 999
```
dagwieers: I wondered why this defaulted to 900, and now to 999.
Wouldn't it be better to take something very large, or -1 or 0 instead?

I don't know exactly how this impacts resumepoints and Kodi (does this value gets stored?).

resources/lib/playerinfo.py
```diff
@@ -198,8 +215,8 @@ def push_position(self, position=0, total=100, event=None):
             asynchronous=True
         )
 
-        # Do not reload container when we are not done playing
-        if event not in ('onPlayBackEnded', 'onPlayBackError', 'onPlayBackStopped'):
+        # Do not reload container when a single episode was played
```
dagwieers: Why? I still do not completely understand our strategy here.
Is this because we do a reload elsewhere as well?

*(I had it disabled because with `onThreadExit()` it appeared to work without reload)*

resources/lib/playerinfo.py
```diff
@@ -178,10 +193,12 @@ def update_total(self):
         ''' Update the total video time '''
         try:
             self.total = self.getTotalTime()
+            if not self.total:
```
dagwieers: We should keep the logic within a `try:`-block to a minimal.
So this ought to go into an `else:`-block after the `except:`-block.

dagwieers: BTW If nothing is being played (i.e. it just finished or was stopped) we get a RuntimeError, so maybe this really ought to go in a `finally:`-block instead?

resources/lib/playerinfo.py
```diff
@@ -117,38 +125,45 @@ def onPlayBackEnded(self):  # pylint: disable=invalid-name
...
         while self.isPlaying() and not self.quit.is_set():
             self.update_position()
             if self.quit.wait(timeout=0.2):
                 break
         self.onThreadExit()
 
+    def add_upnext(self, video_id):
+        ''' Add Up Next url to Kodi Player '''
+        url = 'plugin://plugin.video.vrt.nu/play/upnext/%s' % video_id
+        if self.isPlaying() and self.getTotalTime() - self.getTime() < 1:
```
dagwieers: Beware that because this is not atomic, either getTotalTime() or getTime() can cause a RuntimeError exception. I have seen this happen, which is why I wrote those wrappers.

resources/lib/playerinfo.py
```diff
@@ -72,8 +82,12 @@ def onAVStarted(self):  # pylint: disable=invalid-name
         self.quit.clear()
         self.update_position()
         self.update_total()
+        self.push_upnext()
 
-        Thread(target=self.stream_position, name='StreamPosition').start()
+        # Only start StreamPosition Thread when necessery
```
dagwieers: Small typo.
```suggestion
        # Only start StreamPosition Thread when necessary
```

Also could you elaborate (in comments) when or why this is/isnot  necessary ?

dagwieers: Also, I think the variable name `streampos` is not clear. I assumed this was holding a position, like last_pos.
Not sure about a better name, maybe "positionthread"?

resources/lib/playerinfo.py
```diff
@@ -49,20 +51,28 @@ def onPlayBackStarted(self):  # pylint: disable=invalid-name
...
         self.url = url_to_episode(episode.get('url', ''))
+        self.ep_id = 'S%sE%s' % (episode.get('seasonTitle'), episode.get('episodeNumber'))
         self.whatson_id = episode.get('whatsonId') if episode.get('whatsonId') else None
-
         self.update_position()
-        self.push_position(position=self.last_pos, total=self.total, event='onPlayBackStarted')
+        self.update_total()
+        self.push_position(position=self.last_pos, total=self.total)
+
+    def get_episode_data(self, ep_id):
```
dagwieers: This function might as well be an apihelper method, since it only accesses the ApiHelper object.
Although we may want to indicate in the method-description where/how this is used.

resources/lib/playerinfo.py
```diff
@@ -85,18 +99,13 @@ def onPlayBackSeek(self, time, seekOffset):  # pylint: disable=invalid-name
         log(3, '[PlayerInfo %d] Event onPlayBackSeek time=%d offset=%d' % (self.thread_id, time, seekOffset))
         self.last_pos = time // 1000
 
-        # If we seek beyond the end, quit Player
-        if self.last_pos >= self.total:
```
dagwieers: So the reason to have this, was to ensure that seeking until the end would effectively end the current stream.

I have seen the following behaviour in the past:
- You seek to the end (e.g. to mark the video as watched)
- It would jump back to the last position and continue playing

resources/lib/playerinfo.py
mediaminister: You added `push_position` in `onPlayBackStarted` but at that time, there is no video info yet and `xbmc.Player().getTotalTime()` returns 0.0
So, the intention was to reset the resumepoint to "unwatched" and sending position 0.0 and total 999 to VRT NU. I'm not sure what is the best choice here.

Sending real values can also be done, but only at a later time in `onAVStarted`.

dagwieers: I did not realize.

If I remember correctly, if you would tag an episode as Watch Later, VRT NU would automatically set resumetime to 0 and totaltime to 100. Which is odd, but at least that is what the VRT NU website is doing.

resources/lib/playerinfo.py
mediaminister: No, in this implementation there is only an extra reload after watching multiple episodes because for some reason Kodi doesn't update the last one of multiple watched episodes.
For watching single episodes, I rely on Kodi internal watch status.

It seems Kodi internal watch status is only updated when the play action is initiated from the GUI, so this only works for single episodes.

dagwieers: Thanks for clarifying. That information is crucial to have as comments in the code.
Not just because I have a terrible memory, but also because we may not be around :-)

resources/lib/playerinfo.py
mediaminister: I did this because `self.getTotalTime()` returns 0.0 in `onPlayBackStarted`

resources/lib/playerinfo.py
mediaminister: I'll use the wrappers.

resources/lib/playerinfo.py
mediaminister: StreamPosition thread keeps running when watching multiple episode with "Up Next", only start StreamPosition thread when it doesn't exist yet.

resources/lib/playerinfo.py
mediaminister: Okay, I'll readd this just to be sure.

resources/lib/playerinfo.py
mediaminister: Okay, I'll change it to what VRT NU does.

resources/lib/playerinfo.py
mediaminister: I'll extend the comment with this information.

resources/lib/playerinfo.py
dagwieers: I understand, but I think we need to do this:
```python
        try:
            self.total = self.getTotalTime()
        except RuntimeError:
            pass
        finally:
            if not self.total:
                self.total = 100
```

resources/lib/playerinfo.py
dagwieers: We could have better/other wrappers that return a value (or None) and handle the exception. 
The assignment can then go back in the real code. I think that will read better.

resources/lib/playerinfo.py
mediaminister: Okay, I'll add this to apihelper.

resources/lib/playerinfo.py
dagwieers: No, let's leave it out. We can put it back if it is needed.

resources/lib/playerinfo.py
dagwieers: Leave it at None. We set it to 100 when we create an entry in ResumePoints.update() already.
At least that way we do not override a better value with 100.

resources/lib/playerinfo.py
mediaminister: I'll remove this.

PR #612, opened 2019-12-01
Make HTTP cache ttl configurable

### LINKED ISSUE #541: Make HTTP cache ttl configurable
### Describe the feature ###
I have always felt that the HTTP cache ttl should be configurable in the settings. The way I see it if people enable HTTP caching we would need two different values/sliders:
- direct cache: 5mins
- indirect cache 60mins

The difference is that for some menu listings we expect the user to want to have a more recent update than for some other menu listings. E.g. the Most recent listing probably requires a more recent cache copy, than the default programs listings (which changes less often). And e.g. the favorites/resumepoints lists are more important in the favorites sub-menu, then they are in the normal menus (where they are only used to indicate what is followed/watched later).

dagwieers:
This makes the HTTP cache ttl configurable rather than hard-coded.

We implemented a 'direct' cache ttl (5min) and an 'indirect' cache
ttl (60min)based on the relation the cache has to the listing that
requires it.

This fixes #541

### DIFF
--- a/resources/settings.xml
+++ b/resources/settings.xml
@@ -60,20 +60,22 @@
         <setting label="30881" help="30882" type="action" action="InstallAddon(script.module.pysocks)" visible="!System.HasAddon(script.module.pysocks)"/>
     </category>
     <category label="30900"> <!-- Expert -->
-        <setting label="30901" type="lsep"/> <!-- Cache -->
-        <setting label="30903" help="30904" type="action" action="RunPlugin(plugin://plugin.video.vrt.nu/tokens/delete)"/> <!-- Delete tokens -->
-        <setting label="30905" help="30906" type="action" action="RunPlugin(plugin://plugin.video.vrt.nu/favorites/refresh)"/> <!-- Refresh favorites -->
-        <setting label="30907" help="30908" type="action" action="RunPlugin(plugin://plugin.video.vrt.nu/resumepoints/refresh)"/> <!-- Refresh resumepoints -->
-        <setting label="30909" help="30910" type="bool" id="usemenucaching" default="true"/>
-        <setting label="30911" help="30912" type="bool" id="usehttpcaching" default="true"/>
-        <setting label="30913" help="30914" type="action" action="RunPlugin(plugin://plugin.video.vrt.nu/cache/delete)" enable="eq(-1,true)" subsetting="true"/>
-        <setting label="30915" type="lsep"/> <!-- InputStream Adaptive -->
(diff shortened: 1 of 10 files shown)


resources/settings.xml
```diff
@@ -59,20 +59,22 @@
...
+        <setting label="30915" help="30916" type="action" action="RunPlugin(plugin://plugin.video.vrt.nu/tokens/delete)"/> <!-- Delete tokens -->
+        <setting label="30917" help="30918" type="action" action="RunPlugin(plugin://plugin.video.vrt.nu/favorites/refresh)"/> <!-- Refresh favorites -->
+        <setting label="30919" help="30920" type="action" action="RunPlugin(plugin://plugin.video.vrt.nu/resumepoints/refresh)"/> <!-- Refresh resumepoints -->
+        <setting label="30921" help="30922" type="bool" id="usemenucaching" default="true"/>
+        <setting label="30923" help="30924" type="bool" id="usehttpcaching" default="true"/>
+        <setting label="30925" help="30926" type="action" action="RunPlugin(plugin://plugin.video.vrt.nu/cache/delete)" enable="eq(-1,true)" subsetting="true"/>
+        <setting label="30927" help="30928" type="slider" id="httpcachettldirect" default="5" range="1,1,480" option="int" enable="eq(-2,true)" subsetting="true"/>
+        <setting label="30929" help="30930" type="slider" id="httpcachettlindirect" default="60" range="1,1,480" option="int" enable="eq(-3,true)" subsetting="true"/>
```
mediaminister: Does it make sense that a user can set both types of cache to 480 minutes?

dagwieers: Probably not, but the only reason to do this is to make both sliders visually comparable. Not sure if that is a must. I haven't fully thought about the use-cases really.

Also, 480 minutes (6 hours) is a questionable max value. Probably 4 hours or maybe 2 hours is sufficient. (Unless people do want very long values, in theory they could specify both ttl values that expires only at midnight) In fact, we probably should expire caches at midnight.

And a per minute step is also questionable for the indirect value.

dagwieers: I turned it down to 240 minutes, if there is ever a use-case to increase it, we can still do it.

I also was wondering why we still needed the option to turn HTTP caching off, as you can set both to 0, but this is easier *and* we do have some caching included that are not tied to 'direct' and 'indirect'.

dagwieers:
Let's merge. We can still modify this before or after releasing if we want to.

PR #567, opened 2019-11-03
Improvements to PlayerInfo

dagwieers:
This PR includes:
- ~Add onPlayBackSeek, onThreadExit and onPlayBackStarted event handlers~ #597 
- ~Asynchronous resumepoints updates~ #595 
- ~Implement window properties functions~ #597 
- ~Store Container.URL in Window properties~ #597 
- Add unique ids to ListItem objects
- ~Simplify Service class and improve PlayerInfo class~ #597

~The onPlayBackSeek and onPlayBackStarted events are required so seeking and the initial start (from beginning) would update to correct timing. If not, quitting immediately after playing or after seeking would not get registered (as the position would not have updated).~
    
~The asynchronous resumepoint update is useful because currently updating resumepoints is a blocking call and could interfere with normal operations (i.e. when having network issues).~
    
~In all other cases updating favorites or resumepoints is already asynchronous because they are initiated by a context menu.~

PR #613, opened 2019-12-02
More PlayerInfo improvements

dagwieers:
This PR includes:
- Catch exception when getting position or totaltime
- Push upnext in thread
- A new container_reload() which only reloads the original menu
- Change conditions of when container needs to be reloaded (disable it seems to work)
- Quit player when we seek to the end (this was broken by something?)

This includes the safer/saner bits from #567

### DIFF
--- a/resources/lib/kodiutils.py
+++ b/resources/lib/kodiutils.py
@@ -789,6 +789,15 @@ def container_update(url=None):
         container_refresh()
 
 
+def container_reload(url=None):
+    ''' Only update container if the play action was initiated from it '''
+    if url is None:
+        url = get_property('container.url')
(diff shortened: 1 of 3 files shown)


dagwieers:
I am probably neglecting something important, but this seems to work without any container reload. So I disabled it for now (in fact, only onThreadExit() updates position, and it does not trigger a reload now). Still the menu has an updated resumetime just fine when we return from the Player.

This is what I have seen before and it appears that Kodi is smart to update the resumetime after returning from Player, without requiring a reload of the container.

dagwieers:
I am going to merge this so this can have more testing, and I can finally reduce #567 to a bare state.

PR #617, opened 2019-12-03
Fix broken unfollow functionality

mediaminister:


dagwieers:
That fix I already added to #605 , but it did not appear to work and I didn't look into it further. Sorry for the havoc.

PR #618, opened 2019-12-03
Update 'Watch later' after watching episode

mediaminister:
This pull request fixes an issue where the current directory listing is not updated after finishing a "Watch later" episode.

As Kodi uses different resumepoint margins than VRT NU and Kodi has no API to read these variables, I hardcoded the standard Kodi margins to force a container reload when needed.

You can find more information about the Kodi resumepoints margins at https://kodi.wiki/view/HOW-TO:Modify_automatic_watch_and_resume_points

### DIFF
--- a/resources/lib/playerinfo.py
+++ b/resources/lib/playerinfo.py
@@ -6,10 +6,11 @@ from __future__ import absolute_import, division, unicode_literals
 from threading import Event, Thread
 from xbmc import getInfoLabel, Player, PlayList
 from apihelper import ApiHelper
+from data import SECONDS_MARGIN
 from favorites import Favorites
 from resumepoints import ResumePoints
 from statichelper import play_url_to_id, to_unicode, url_to_episode
-from kodiutils import addon_id, container_reload, get_setting, has_addon, log, notify
+from kodiutils import addon_id, container_reload, get_advanced_setting, get_setting, has_addon, log, notify
 
 
 class PlayerInfo(Player):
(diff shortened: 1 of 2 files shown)


resources/lib/playerinfo.py
```diff
@@ -201,20 +202,35 @@ def push_position(self, position=0, total=100):
...
         )
 
-        # Do not reload container and rely on Kodi internal watch status, when watching a single episode.
+        # Kodi uses different resumepoint margins than VRT NU, to obey to VRT NU resumepoint margins we need to reload the container
+        # to overwrite Kodi watch status when the position is in these margins.
+        # Use setting from advancedsettings.xml or default value
+        # https://github.com/xbmc/xbmc/blob/master/xbmc/settings/AdvancedSettings.cpp
+        # https://kodi.wiki/view/HOW-TO:Modify_automatic_watch_and_resume_points
+        ignoresecondsatstart = get_advanced_setting('video/ignoresecondsatstart') or 180
+        ignorepercentatend = get_advanced_setting('video/ignorepercentatend') or 8
```
dagwieers: Beware that if people have configured this to be 0, it would go to the default.

It is probably better to have this interface: `def get_advanced_setting(key, default=None):`
And then call it as:
```python
        ignoresecondsatstart = get_advanced_setting('video/ignoresecondsatstart', default=180)
        ignorepercentatend = get_advanced_setting('video/ignorepercentatend', default=8)
```

mediaminister: Fixed.

dagwieers:
My only concern here is that this information is from advancedsettings, which is a feature that should not exist IMO.

mediaminister:
This information is only used to limit the number of container refreshes to a minimum. (We rely on Kodi internal watch status when possible).
We can also ignore this information, but this increases the number of container refreshes.

dagwieers:
I know, it's more a criticism of Kodi. This should have been part of the normal Kodi settings.

PR #619, opened 2019-12-07
Use unicode with json module

mediaminister:
While testing I experienced that passing a file handle from `xbmcvfs` to `json` module doesn't work very reliable. Sometimes it fails to read the file properly which causes a `ValueError` on Python2 or a `JSONDecodeError` on Python3.
When testing the json file manually, there is never a problem, it's always valid json.

Directly passing unicode to the `json` module seems to fix these errors.

### DIFF
--- a/resources/lib/kodiutils.py
+++ b/resources/lib/kodiutils.py
@@ -697,18 +697,13 @@ def get_cache(path, ttl=None):  # pylint: disable=redefined-outer-name
     mtime = stat_file(fullpath).st_mtime()
     now = mktime(localtime())
     if ttl is None or now - mtime < ttl:
-        from json import load
+        from json import loads
         if ttl is None:
             log(3, "Cache '{path}' is forced from cache.", path=path)
         else:
             log(3, "Cache '{path}' is fresh, expires in {time}.", path=path, time=human_delta(mtime + ttl - now))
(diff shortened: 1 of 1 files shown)


dagwieers:
But the TypeError and ValueError exception are required for the cases where you don't get valid JSON when you expect it. Usually this is because of back-end issues or I/O issues.

mediaminister:
I'll readd a try catch structure that also takes into account Python3 `JSONDecodeError`.

dagwieers:
And an abstraction layer?

mediaminister:
Not yet, I first have to find the root cause of the file reading problem, I consider this a workaround.

dagwieers:
Even better, because I haven't seen this myself. The TypeErrors and ValueErrors I encountered were al related to network-type issues. If we can hand off http requests to our own http library, we can properly intercept HTTP error codes, implement retries and expose sensible errors indicating possible causes. Something for v2.4.0 maybe?

dagwieers:
Oh wait, wad this causing my empty directory listings when returning to a menu 🙄

mediaminister:
This is very possible, when I remove the `try...except` block in `get_cache` I regularly get errors while reading `resume_points.json` from cache.

PR #621, opened 2019-12-07
Catch json errors

mediaminister:


mediaminister:
Maybe in the future, I should first understand why reading json files in Kodi fails so many times.

PR #622, opened 2019-12-08
Fix ValueError undefined on Python 3

mediaminister:


dagwieers:
This is the wrong fix, as it would catch any exception.

The solution is to make your own class inheriting from Exception.

```
class ValueError(Exception):
    pass
```

PR #623, opened 2019-12-08
Make exceptions Python 2/3 compatible

mediaminister:


resources/lib/kodiutils.py
```diff
@@ -699,10 +708,10 @@ def get_cache(path, ttl=None):  # pylint: disable=redefined-outer-name
     if ttl is None or now - mtime < ttl:
         try:  # Python 3
             from json import loads, JSONDecodeError
-            ValueError = BaseException  # pylint: disable=invalid-name, redefined-builtin
+            ValueError = StubException()  # pylint: disable=invalid-name, redefined-builtin
```
dagwieers: I don't think you have to instantiate exceptions.

resources/lib/kodiutils.py
```diff
@@ -9,6 +9,15 @@
 import xbmcplugin
 from statichelper import from_unicode, to_unicode
 
+
+class StubException(Exception):
```
dagwieers: I would rather define it like this:
```python
        try:  # Python 3
            from json import loads, JSONDecodeError
            class ValueError(Error):
                pass
         except ImportError:  # Python 2
            from json import loads
            class JSONDecodeError(Error):
                pass
```

resources/lib/kodiutils.py
dagwieers: In fact I think this is wrong, because you catch Exception or Error classes not instances.

mediaminister: In Python 3 I got:
`TypeError: catching classes that do not inherit from BaseException is not allowed`

dagwieers: In that case you may have to do: `class ValueError(Error, BaseException):` ?

dagwieers: Ah, Error does not even exist. My bad. `class ValueError(BaseException):`

resources/lib/kodiutils.py
dagwieers: BTW I think ValueError also exists in Python 3, so I don't think we have to define it here.

dagwieers:
You also may want to read this:
https://stackoverflow.com/questions/15170676/cannot-import-jsondecodeerror

Since JSONDecodeError is a subclass of ValueError, you don't need to catch that specifically.

So we only need what we used to have and JSONDecodeError exceptions are caught as ValueError.
```python
    from json import loads

    try:
        ...
    except (TypeError, ValueError):
        ...
```

mediaminister:
> So we only need what we used to have.

You're probably right. I'm sorry for messing up, I can't reproduce any of the reading errors I got yesterday.

dagwieers:
I didn't know this either. And would probably have done the same thing, adding the exceptions that are raised. Learned something new today!

resources/lib/kodiutils.py
```diff
@@ -697,24 +697,19 @@ def get_cache(path, ttl=None):  # pylint: disable=redefined-outer-name
...
         if ttl is None:
             log(3, "Cache '{path}' is forced from cache.", path=path)
         else:
             log(3, "Cache '{path}' is fresh, expires in {time}.", path=path, time=human_delta(mtime + ttl - now))
         cache_data = None
         with open_file(fullpath, 'r') as fdesc:
-            try:
-                cache_data = to_unicode(fdesc.read())
-                if cache_data:
+            cache_data = to_unicode(fdesc.read())
```
dagwieers: I don't know if you intended to also catch exceptions from to_unicode() or read() as well.
If so, we should use a different try-except block.

It's always best to use different try-blocks for each statement.

mediaminister: No, exceptions only occur when passing the cache_data to the json module.
```
2019-12-08 22:32:19.050 T:140696274335488   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                            Error Type: <type 'exceptions.ValueError'>
                                            Error Contents: Expecting object: line 1 column 94528 (char 94527)
                                              File "/home/test/.kodi/addons/plugin.video.vrt.nu/resources/lib/addon_entry.py", line 14, in <module>
                                                run(argv)
                                              File "/home/test/.kodi/addons/plugin.video.vrt.nu/resources/lib/addon.py", line 298, in run
                                                plugin.run(argv)
                                              File "/home/test/.kodi/addons/script.module.routing/lib/routing.py", line 130, in run
                                                self._dispatch(self.path)
                                              File "/home/test/.kodi/addons/script.module.routing/lib/routing.py", line 141, in _dispatch
                                                view_func(**kwargs)
                                              File "/home/test/.kodi/addons/plugin.video.vrt.nu/resources/lib/addon.py", line 140, in resumepoints_watchlater
                                                VRTPlayer().show_watchlater_menu(page=1)
                                              File "/home/test/.kodi/addons/plugin.video.vrt.nu/resources/lib/vrtplayer.py", line 303, in show_watchlater_menu
                                                self._resumepoints.refresh(ttl=ttl('direct'))
                                              File "/home/test/.kodi/addons/plugin.video.vrt.nu/resources/lib/resumepoints.py", line 37, in refresh
                                                resumepoints_json = get_cache('resume_points.json', ttl)
                                              File "/home/test/.kodi/addons/plugin.video.vrt.nu/resources/lib/kodiutils.py", line 707, in get_cache
                                                return load(fdesc)
                                              File "/usr/lib/python2.7/json/__init__.py", line 291, in load
                                                **kw)
                                              File "/usr/lib/python2.7/json/__init__.py", line 339, in loads
                                                return _default_decoder.decode(s)
                                              File "/usr/lib/python2.7/json/decoder.py", line 364, in decode
                                                obj, end = self.raw_decode(s, idx=_w(s, 0).end())
                                              File "/usr/lib/python2.7/json/decoder.py", line 380, in raw_decode
                                                obj, end = self.scan_once(s, idx)
                                            ValueError: Expecting object: line 1 column 94528 (char 94527)

```

This exception happens after playing a video for more than 5 minutes and stopping the video.

dagwieers: Hmmm, that is weird.
We ought to dump that content :-/

dagwieers: Maybe we are having issues with `update_cache()`, `json.dump()` and encoding?

dagwieers:
Going to merge this, we can improve on this further when we find the true cause.

dagwieers:
I just dumped that resumepoints.json that gave a JSONDecodeError and it was cut off at 4096 bytes.

dagwieers:
Maybe the easiest solution is to drop support for Python 3.5 in the VRT NU add-on.
Because of this Python 3.5 issue:
```
Error: [plugin.video.vrt.nu] apihelper.get_tvshows: the JSON object must be str, not 'bytes'
```

The alternative is that we implement an abstraction layer in v2.3.0 that handles this properly. (i.e. we use a different logic for Python 3.5 specifically)

resources/lib/kodiutils.py
mediaminister: No, I tested and `resume_points.json` on disk is always a valid json file.

PR #628, opened 2019-12-11
Use ActivateWindow to update container

mediaminister:
It seems this works better than `Container.Update`: https://github.com/pietje666/plugin.video.vrt.nu/issues/616

### DIFF
--- a/resources/lib/kodiutils.py
+++ b/resources/lib/kodiutils.py
@@ -694,8 +694,11 @@ def container_refresh(url=None):
 def container_update(url=None):
     ''' Update the current container with respect for the path history. '''
     if url:
-        log(3, 'Execute: Container.Update({url})', url=url)
-        xbmc.executebuiltin('Container.Update({url})'.format(url=url))
+        # log(3, 'Execute: Container.Update({url})', url=url)
+        # xbmc.executebuiltin('Container.Update({url})'.format(url=url))
+        # ActivateWindow seems to work better after closing Kodi Player
+        log(3, 'Execute: ActivateWindow(Videos,{url})', url=url)
+        xbmc.executebuiltin('ActivateWindow(Videos,{url})'.format(url=url))
     else:
         # URL is a mandatory argument for Container.Update, use Container.Refresh instead
         container_refresh()


resources/lib/kodiutils.py
```diff
@@ -694,8 +694,11 @@ def container_refresh(url=None):
 def container_update(url=None):
     ''' Update the current container with respect for the path history. '''
     if url:
-        log(3, 'Execute: Container.Update({url})', url=url)
-        xbmc.executebuiltin('Container.Update({url})'.format(url=url))
+        # log(3, 'Execute: Container.Update({url})', url=url)
+        # xbmc.executebuiltin('Container.Update({url})'.format(url=url))
+        # ActivateWindow seems to work better after closing Kodi Player
+        log(3, 'Execute: ActivateWindow(Videos,{url})', url=url)
+        xbmc.executebuiltin('ActivateWindow(Videos,{url})'.format(url=url))
```
dagwieers: Should this not use `ReplaceWindow()` instead? Let me test...

dagwieers: Also, I fear that if the container has changed, the user may suddenly end up in a different container. So in this case the protection better works 100% !

dagwieers: The problem is with all our improvements I cannot make this do what we expect it to.
It never gets triggered, the Kodi container refresh always works correctly now.

dagwieers: Hmm, after testing this using `kodi-send` this is not doing anything if the URL provided is exactly the same as the virtual directory the we have open.

In other words, it does the opposite of what we want. We want to refresh the container only when it is the exact same URL as is currently active.

dagwieers: Also, there appears no difference in `ActivateWindow()` and `ReplaceWindow()`.
According to the documentation `ActivateWindow()` adds a new window to the history, and `ReplaceWindow()` does not. The effect of both is that if you go back, you go to the parent of the parent. So you are replacing the parent in the history, not the current entry. :disappointed:

mediaminister: I see, I'll close this.

PR #629, opened 2019-12-11
Prevent Kodi Player errors

mediaminister:
We can prevent RuntimeErrors by checking if Kodi Player is playing.

### DIFF
--- a/resources/lib/playerinfo.py
+++ b/resources/lib/playerinfo.py
@@ -171,7 +171,7 @@ class PlayerInfo(Player):
 
     def push_upnext(self):
         ''' Push episode info to Up Next service add-on'''
-        if has_addon('service.upnext') and get_setting('useupnext', 'true') == 'true':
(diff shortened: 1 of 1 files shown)


resources/lib/playerinfo.py
```diff
@@ -189,17 +189,15 @@ def push_upnext(self):
 
     def update_position(self):
         ''' Update the player position, when possible '''
-        try:
-            self.last_pos = self.getTime()
-        except RuntimeError:
-            pass
+        if not self.isPlaying():
+            return
+        self.last_pos = self.getTime()
```
dagwieers: This is not correct. It may be playing when you check, and it may have stopped when you get the position.

The probability that this happens is very small, but it can still happen. Unfortunately, exception handling in this case is the only real option. (and yes, almost every other add-on I have seen are doing this wrong as well)

mediaminister: Okay, I'll close this.

PR #632, opened 2019-12-11
Add UpNext play_url support

dagwieers:
This adds UpNext `play_url` support and removes the old interface.

The aim is to move this functionality to Up Next so it is easier to
integrate with Up Next and offer the best experience in Up Next.
(Without work-arounds in VRT NU)

This would require play_url support in Up Next.


dagwieers:
@michaelarnauts Now I get both Codecov and Sonarcloud to send comments :roll_eyes:

michaelarnauts:
It is beginning. The bots are taking over!

dagwieers:
You can disable this per repository: https://jira.sonarsource.com/browse/SCCOMM-19

But it is enabled by default :-( And we cannot disable it for the whole organization...

michaelarnauts:
Out of curiousity, is there anything blocking on this?

dagwieers:
@michaelarnauts Because our own handling is much more sophisticated than what is currently implemented in Up Next. But I think we need to re-evaluate Kodi's behaviour so we don't need anything special at all.

dagwieers:
I reevaluated this with the recent changes to Up Next, and it does seem to work well now. You now get a progress circle running on top of the running episode until the new episode is loaded. It does feel to take a tiny bit longer than before.

PR #633, opened 2019-12-11
Extend jsonrpc() to support list of commands

dagwieers:
This extends the existing interface to support more than one command.

So you can now do any of the below:
```python
jsonrpc(method='Input.Down')  # Use kwargs (as used now)
jsonrpc(dict(method='Input.Down'))  # Use dict as arg
jsonrpc(dict(method='Input.Down'), dict(method='Input.Up'))  # Use dicts as args
```


dagwieers:
Still need to test this against Kodi to be sure this is in fact supported as seen in examples on forums.

PR #636, opened 2019-12-12
Reduce manual container updates to absolute minimum

mediaminister:
There were several mistakes in the logic to determine if the container should be updated manually. I rewrote it to reduce container updates even more.

### DIFF
--- a/resources/lib/playerinfo.py
+++ b/resources/lib/playerinfo.py
@@ -220,24 +220,47 @@ class PlayerInfo(Player):
             asynchronous=True
         )
 
-        # Kodi uses different resumepoint margins than VRT NU, to obey to VRT NU resumepoint margins we need to reload the container
-        # to overwrite Kodi watch status when the position is in these margins.
-        # Use setting from advancedsettings.xml or default value
-        # https://github.com/xbmc/xbmc/blob/master/xbmc/settings/AdvancedSettings.cpp
-        # https://kodi.wiki/view/HOW-TO:Modify_automatic_watch_and_resume_points
-        ignoresecondsatstart = get_advanced_setting('video/ignoresecondsatstart', default=180)
-        ignorepercentatend = get_advanced_setting('video/ignorepercentatend', default=8)
-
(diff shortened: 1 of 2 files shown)


dagwieers:
Should we have unit tests for this?

mediaminister:
> Should we have unit tests for this?

I added a simple test.

resources/lib/playerinfo.py
```diff
@@ -220,24 +220,47 @@ def push_position(self, position=0, total=100):
...
+
+        ignoresecondsatstart = get_advanced_setting('video/ignoresecondsatstart', default=180)
+        ignorepercentatend = get_advanced_setting('video/ignorepercentatend', default=8)
+
+        # Convert percentage to seconds
+        ignoresecondsatend = round(total - (100 - ignorepercentatend) / 100.0 * total)
+
+        if position <= max(SECONDS_MARGIN, ignoresecondsatstart):
+            # Check start margins
+            if ignoresecondsatstart >= SECONDS_MARGIN:
```
dagwieers: Would this not be identical:
```python
            if SECONDS_MARGIN <= position <= ignoresecondsatstart:
                return True
            elif ignoresecondsatstart <= position <= SECONDS_MARGIN:
                return True
```

resources/lib/playerinfo.py
```diff
@@ -220,24 +220,47 @@ def push_position(self, position=0, total=100):
...
+                if SECONDS_MARGIN <= position <= ignoresecondsatstart:
+                    return True
+            elif ignoresecondsatstart <= position <= SECONDS_MARGIN:
+                return True
+
+        if position >= min(total - SECONDS_MARGIN, total - ignoresecondsatend):
+            # Check end margins
+            if ignoresecondsatend >= SECONDS_MARGIN:
```
dagwieers: Would this also not be identical:
```python
            if SECONDS_MARGIN <= total - position <= ignoresecondsatend:
                return True
            elif ignoresecondsatend <= total - position <= SECONDS_MARGIN:
                return True
```

PR #643, opened 2019-12-16
Refresh Continue Watching menu when deleting a resumepoint

mediaminister:


resources/lib/resumepoints.py
```diff
@@ -183,6 +185,12 @@ def refresh_watchlater():
         invalidate_caches('watchlater-*.json')
         container_refresh()
 
+    @staticmethod
+    def refresh_continuewatching():
+        ''' Refresh Continue Watching menu '''
+        invalidate_caches('continue-*.json')
```
dagwieers: That was biting me as well it appears. Good catch.

resources/lib/resumepoints.py
```diff
@@ -183,6 +185,12 @@ def refresh_watchlater():
...
+        container_refresh()
```
dagwieers: Did you test without this?
I wonder if a container_refresh() is canceled when it is scheduled before Kodi is doing one anyway.
Or whether it always causes a refresh even when not needed.

mediaminister: Yes, I added this after testing. But maybe only calling `invalidate_caches` is sufficient to make this work.

dagwieers: It all depends on timing I think. And we made a lot of improvements on being more efficient in those codepaths. If the forced refresh doesn't induce a second refresh this would be actually a good thing (if it doesn't get delayed), but if it does we may want to look for something better eventually.

In any case, it is better to have an additional refresh when not needed, than no refresh when needed.

PR #645, opened 2019-12-16
Don't update total to 0.0

mediaminister:
Another Kodi bug that came to surface, Kodi Player sometimes returns a total time of 0.0 and this causes unwanted behaviour with VRT NU Resumepoints API.
We should ignore this zero value and pass the previous value or the initial `None` value to VRT NU Resumepoints API.

### DIFF
--- a/resources/lib/playerinfo.py
+++ b/resources/lib/playerinfo.py
@@ -193,7 +193,10 @@ class PlayerInfo(Player):
     def update_total(self):
         ''' Update the total video time '''
         try:
-            self.total = self.getTotalTime()
+            total = self.getTotalTime()
+            # Kodi Player sometimes returns a total time of 0.0 and this causes unwanted behaviour with VRT NU Resumepoints API.
+            if total > 0.0:
+                self.total = total
         except RuntimeError:
             pass
 


dagwieers:
I guess the total time should never be 0 anyway :-) Another good catch.

It sometimes wonders me if we should have a *very verbose* mode that logs everything under the sun for weird cases like this.

And should we report this to Kodi, or is this an expected value for when e.g. Player is not (yet/no longer) playing?

resources/lib/playerinfo.py
```diff
@@ -193,7 +193,10 @@ def update_position(self):
     def update_total(self):
         ''' Update the total video time '''
         try:
-            self.total = self.getTotalTime()
+            total = self.getTotalTime()
+            # Kodi Player sometimes returns a total time of 0.0 and this causes unwanted behaviour with VRT NU Resumepoints API.
+            if total > 0.0:
```
dagwieers: Good you are anticipating negative total times here too. Who know what will happen in the future ;-)

dagwieers:
Merging because I want to test the whole lot.

mediaminister:
The only thing I currently know is that we don't want zero or lower values in our implementation.
Better debugging options is something to think about in 2020 for sure.

PR #647, opened 2019-12-16
Assorted fixed

dagwieers:
This PR includes:
- Small fixes to .gitignore and Makefile
- Make use of triple double quotes for docstrings (PEP257)
  - Don't like it, but hey, it's another convention for our own benefit!
- Support custom domains in reformat_url()
- Add `JSONRPC.NotifyAll` support, including a working implementation from Addon Signals
- Add `xbmcvfs.copy` support

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,8 +71,7 @@ msbuild.wrn
 .vscode
 
 # Test caches
-*.pyc
-*.pyo
+*.py[cod]
 .coverage
 .pytest_cache/
 .tox/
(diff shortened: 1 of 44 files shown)


dagwieers:
@mediaminister Let me know if you agree to this.

The change to `reformat_url()` is to assist with some scripts I have been developing that are using our interface.

dagwieers:
Also, the implementation for `JSONRPC.NotifyAll` opens up a whole set of possibilities for a potential future xbmc.Player implementation. Which would be quite useful for UpNext as well.

mediaminister:
Please merge.

PR #650, opened 2019-12-17
Assorted resumepoint fixes

mediaminister:
This pull request includes:
- Avoid setting resumepoints for livestreams
- Avoid updating resumepoints with data from previous episode
- Avoid sending zero position values because this triggers unwanted resumepoint deletion
- Move local resumepoint cache updates to separate functions
- Use `update_online` json response to update local cache
- Make sure unit tests don't remove existing watchLater episodes or permanently add new watchLater episodes

### DIFF
--- a/resources/lib/playerinfo.py
+++ b/resources/lib/playerinfo.py
@@ -7,7 +7,7 @@ from threading import Event, Thread
 from xbmc import getInfoLabel, Player, PlayList
 
 from apihelper import ApiHelper
-from data import SECONDS_MARGIN
+from data import SECONDS_MARGIN, CHANNELS
 from favorites import Favorites
 from kodiutils import addon_id, container_reload, get_advanced_setting_int, get_setting, has_addon, log, notify
 from resumepoints import ResumePoints
@@ -24,7 +24,7 @@ class PlayerInfo(Player, object):  # pylint: disable=useless-object-inheritance
         self.last_pos = None
         self.listen = False
         self.paused = False
-        self.total = None
+        self.total = 100
         self.positionthread = None
         self.quit = Event()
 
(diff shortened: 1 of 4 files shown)


resources/lib/playerinfo.py
```diff
@@ -50,21 +50,31 @@ def onPlayBackStarted(self):  # pylint: disable=invalid-name
...
+        # Avoid setting resumepoints without episode data
         if episode is None:
             return
 
         self.asset_id = assetpath_to_id(episode.get('assetPath'))
         self.title = episode.get('program')
         self.url = url_to_episode(episode.get('url', ''))
-        self.ep_id = 'S%sE%s' % (episode.get('seasonTitle'), episode.get('episodeNumber'))
-        self.whatson_id = episode.get('whatsonId') if episode.get('whatsonId') else None
+        self.whatson_id = episode.get('whatsonId') if episode.get('whatsonId') else None  # Avoid empty string
```
dagwieers: This could be written as:
```python
self.whatson_id = episode.get('whatsonId') or None
```

resources/lib/playerinfo.py
```diff
@@ -74,7 +84,7 @@ def onAVStarted(self):  # pylint: disable=invalid-name
         self.quit.clear()
         self.update_position()
         self.update_total()
-        self.push_position()
+        self.push_position(position=self.last_pos, total=self.total)
```
dagwieers: I think we need to add a comment here, because it may not be trivial to others why we update the position at the start.

resources/lib/resumepoints.py
```diff
@@ -134,13 +133,32 @@ def update_online(self, asset_id, title, url, payload):
...
-            get_url_json('https://video-user-data.vrt.be/resume_points/%s' % asset_id, headers=self.resumepoint_headers(url), data=dumps(payload).encode())
+            resumepoint_json = get_url_json('https://video-user-data.vrt.be/resume_points/%s' % asset_id,
+                                            headers=self.resumepoint_headers(url), data=dumps(payload).encode())
         except HTTPError as exc:
             log_error('Failed to (un)watch episode {title} at VRT NU ({error})', title=title, error=exc)
             notification(message=localize(30977, title=title))
             return False
+
+        # Update local representation and cache
+        self.update_local(asset_id, resumepoint_json)
```
dagwieers: Hmm, I was confused here at first, because our internal representation should be identical to what we just pushed.

But I understand that pushing returns a complete representation of the resumepoints online. Just not entirely sure how this will work with concurrency.

mediaminister: I'll remove this for now. Maybe this is something that can be useful in a next pull request.

resources/lib/playerinfo.py
mediaminister: Done.

resources/lib/playerinfo.py
mediaminister: Done.

resources/lib/playerinfo.py
dagwieers: No, it could still be an empty string.

michaelarnauts: Right!

PR #652, opened 2019-12-19
Improve resumepoints

mediaminister:
This pull request includes:
- Reorganize resumepoints.py
- Wait for the resumepoints to be updated before loading the menus

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -875,7 +875,7 @@ msgid "Could not retrieve this program list. VRT Search API url is too long. Unf
 msgstr ""
 
 msgctxt "#30975"
-msgid "Failed to get favorites token from VRT NU"
+msgid "Failed to get user token from VRT NU"
 msgstr ""
 
 msgctxt "#30976"
--- a/resources/lib/playerinfo.py
+++ b/resources/lib/playerinfo.py
@@ -7,9 +7,9 @@ from threading import Event, Thread
 from xbmc import getInfoLabel, Player, PlayList
 
 from apihelper import ApiHelper
-from data import SECONDS_MARGIN, CHANNELS
+from data import CHANNELS
 from favorites import Favorites
-from kodiutils import addon_id, container_reload, get_advanced_setting_int, get_setting, has_addon, log, notify
+from kodiutils import addon_id, container_reload, get_setting, has_addon, log, notify
 from resumepoints import ResumePoints
 from utils import assetpath_to_id, play_url_to_id, to_unicode, url_to_episode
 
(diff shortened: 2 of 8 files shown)


resources/language/resource.language.en_gb/strings.po
```diff
@@ -773,34 +773,38 @@ msgid "Refresh watching activity"
 msgstr ""
 
 msgctxt "#30921"
+msgid "Remove watching activity"
```
dagwieers: I am a bit confused about this new option.
When would the user remove the watching activity, and not disable the watching activity at the same time?
Because as long as the watching activity is enabled, it will be recreated by most listings.

So maybe we ought to remove it if "watching activity" is disabled and the file exists.

dagwieers: Oh, I see that this is emptying everything? Do we need this?
What I proposed to add is an option to clean it up:
- Remove entries that are no longer offered by VRT NU
- Remove entries of which the resume position is not within expected range (e.g. not between 30 and -30)
- Remove entries that have no resume position set, and has watchLater == False

mediaminister: No problem to remove this again.
I mainly added this because the add-on can't update resumepoints for programs like "Het goeie leven" (https://www.vrt.be/vrtnu/a-z/het-goeie-leven/) that don't have an `assetPath` in VRT NU Search API.
There are use cases where these programs are watched on VRT NU website and appear in the Continue and Watch later lists in the add-on, but can't be removed.

dagwieers: How can they appear in the Continue and Watch later lists if they cannot be added to resumepoints?

mediaminister: You can watch these episodes on VRT NU website...and this info is synced with the add-on.

mediaminister: I have opened a separate pull request for this feature: https://github.com/pietje666/plugin.video.vrt.nu/pull/653

resources/lib/playerinfo.py
```diff
@@ -210,62 +217,24 @@ def update_total(self):
...
-        if position >= min(total - SECONDS_MARGIN, ignoresecondsatend):
-            # Check end margins
-            if total - SECONDS_MARGIN <= position <= ignoresecondsatend:
-                return True
-            if ignoresecondsatend <= position <= total - SECONDS_MARGIN:
-                return True
-
-        return False
+        # Set property to let wait_for_resumepoints function know that update resume is done
+        set_property('resumepoints', 'ready')
```
dagwieers: So I don't understand why we wait for the online update, that is a liability IMO.
Only the local resumepoint updates should matter as that is freshly cached.

dagwieers:
If this is ready, it can be merged. The implementation is a lot better.

My only concern is one that other add-ons do not make but is related to the use of properties on Window 10000. IMO it may be better to use a VRT NU related Window instead rather than the Main window. But I have not looked into the details myself yet.

The use of Windows ID 10000 could lead to conflicts with other add-ons (and could require namespacing our properties with e.g. vrtnu_), on the other hand using a VRT NU specific Window ID could potentially lose its information when switching to Kodi windows. (I don't know how Kodi behaves in this regard)

But this is a minor concern given that all add-ons I know are using Windows ID 10000.

mediaminister:
I added `vrtnu_` namespacing to the resumepoints properties. I'm going to merge this.

PR #653, opened 2019-12-21
Add option to completely remove VRT NU resumepoints

mediaminister:
The add-on can't update resumepoints for programs like "Het goeie leven" (https://www.vrt.be/vrtnu/a-z/het-goeie-leven/) that don't have an `assetPath` in VRT NU Search API.
When users watch these programs on the VRT NU **website** and continue watching at a later point in the **add-on**, these programs are not updated and never removed from the add-on.

Currently, this feature removes all resumepoints! Suggested changes or improvements:
- Only remove entries that are no longer offered by VRT NU
- Only remove entries of which the resume position is not within expected range (e.g. not between 30 and -30) and has `watchLater` set to False

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -773,34 +773,38 @@ msgid "Refresh watching activity"
 msgstr ""
 
 msgctxt "#30921"
-msgid "Use menu caching"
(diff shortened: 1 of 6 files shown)


dagwieers:
Thanks for opening this.

PR #659, opened 2020-01-01
Prepare 2.3.0 release again

dagwieers:
Adjust dates.


dagwieers:
The unit test for listing `het-journaal` returns only 3 items and our logic now sorted by `label`, instead of `dateadded`. Must be a Y2K20 thing? :laughing:

PR #669, opened 2020-01-09
Fix livestreams

mediaminister:
VRT removed the live stream player from the channel pages of Eén and Canvas (https://www.vrt.be/vrtnu/kanalen/een/ https://www.vrt.be/vrtnu/kanalen/canvas/) and this broke web scraping live stream attributes for Eén and Canvas.
This pull request fixes this.

This was not really a problem for existing add-on functionality, because the add-on relies on `live_stream_id` in `data.py`, but it's good to keep web scraping up to date as a fallback.

### DIFF
--- a/resources/lib/data.py
+++ b/resources/lib/data.py
@@ -40,7 +40,7 @@ CHANNELS = [
         name='een',
         label='Eén',
         studio='Een',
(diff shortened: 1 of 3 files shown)


dagwieers:
Perfect. Thanks!

I am reorganizing our testing framework a bit, hope to make a PR today.

PR #678, opened 2020-01-16
Check if categories contain all necessary info

mediaminister:
The previous check `if not categories` didn't detect when `name` or `id` was empty and this implies that in some cases the fallback to the internal hard-coded categories doesn't work in v2.3.1 and earlier.

When the cache expires all users will get a blanco categories listing, all new users got already a blanco categories listing when opening the categories menu:

Screenshot at 2020-01-16 20-56-39

We need a 2.3.2 release as soon as possible.


dagwieers:
Oh damn... Well, so be it.

PR #686, opened 2020-02-01
Improve tv guide range: 3 weeks past, 2 weeks future

dagwieers:
Apparently the range for the EPG has changed a bit.

The downside of this is that **Today** is not longer visible from the listing. I would prefer a better view for this.

### DIFF
--- a/resources/lib/tvguide.py
+++ b/resources/lib/tvguide.py
@@ -69,7 +69,7 @@ class TVGuide:
         if epg.hour < 6:
             epg += timedelta(days=-1)
         date_items = []
-        for offset in range(7, -30, -1):
+        for offset in range(14, -19, -1):
             day = epg + timedelta(days=offset)
             label = localize_datelong(day)
             date = day.strftime('%Y-%m-%d')
--- a/test/test_tvguide.py
+++ b/test/test_tvguide.py
@@ -31,11 +31,11 @@ class TestTVGuide(unittest.TestCase):
     def test_tvguide_date_menu(self):
         """Test TV guide main menu"""
         date_items = self._tvguide.get_date_items()
-        self.assertEqual(len(date_items), 37)
+        self.assertEqual(len(date_items), 33)
         date_item = random.choice(date_items)
         print('- %s%s' % (kodi_to_ansi(date_item.label), uri_to_path(date_item.path)))
         date_items = self._tvguide.get_date_items('today')
-        self.assertEqual(len(date_items), 37)
+        self.assertEqual(len(date_items), 33)
         date_item = random.choice(date_items)
         print('- %s%s' % (kodi_to_ansi(date_item.label), uri_to_path(date_item.path)))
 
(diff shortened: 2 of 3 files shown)


michaelarnauts:
Isn't it possible to mark the past days as watched or something, so Kodi highlights today by automatically? (Like it does with series where it selects your next episode by default).

Not sure if this will also work for folders...

dagwieers:
If jumping to a specific item would work in Kodi, we could jump to **Today** by default, unfortunately it doesn't work in my latest testing.

dagwieers:
@michaelarnauts I prefer the future to be on top and the past at the bottom. It is more likely a user needs the future tv guide.

dagwieers:
In fact thinking about this, Kodi should support unplayable ListItems better. Currently we have a nop-action defined for non-playable items so it does no harm clicking on them and we give them another color. But in fact Kodi should theme it, and make it not playable from Kodi.

Non-playable items make sense also for a Series-listing where some episode are not (or no longer) available, but you may have metadata for it anyway (e.g. from EPG).

michaelarnauts:
> @michaelarnauts I prefer the future to be on top and the past at the bottom. It is more likely a user needs the future tv guide.

Actually, most of the programs of the future are not playable (some are if the whole series is put online at once, like Dertigers or so, but that's more of an exception), so I use the guide always to see past programs (the primetime program of yesterday for example).

dagwieers:
@michaelarnauts Obviously today's guide (and the close surroundings, yesterday and tomorrow) are the most important. But we cannot solve that by changing the order. In fact, it would make scrolling down to **Today** even harder if we would turn it around. (We still have a larger past than future).

dagwieers:
The solution will have to come from elsewhere (a different EPG view) or forcing the cursor to be on **Today**.

michaelarnauts:
Or something like:
```
[ Yesterday ] 
[ Today ] 
[ Tomorrow ] 
[ --------------------------------------------------------------------------] 
[ 14/01/2020 ] 
[ 15/01/2020 ] 
[ 16/01/2020 ] 
[ 17/01/2020 ] 
[ 18/01/2020 ] 
...
```

dagwieers:
@michaelarnauts Yes, that would be the next best thing. And this is easily implemented.

dagwieers:
Adding a separator in listings would also be a nice feature to have.
So some sort of separator ListItem that would be rendered as a small line.

michaelarnauts:
Regarding browsing the guide, I load them into the PVR module of Kodi, so I use the Kodi PVR Guide to see future programs (and record them with tvheadend etc). For me, the guide in vrt nu addon is used to see past programs that I haven't recorded.

I can't tell about the "general use case", and I also don't have the average setup :)

dagwieers:
@michaelarnauts So the PVR Guide view would be one way to make this work as well. But I am not sure if this is something that most people would prefer. So if we include it, we still need to have the option which view the user prefers.

PS Whether you can use the PVR EPG view without PVR is the question, I would think it is possible, but haven't investigated ;-)

PR #702, opened 2020-03-23
Use GitHub actions for CI testing

dagwieers:
This PR includes:
- Add a GitHub Actions workflow for CI testing
- Remove Travis CI support
- Move test/ to tests/ (unittest standard?)
- Fix a coloring issue in xbmcextra
- Various new fixes to tracebacks in GitHub CI


dagwieers:
There is one issue with running this in GitHub CI that we can fix later. There is a weird issue related to running **proxy.py** on **Python 3.6.7**. We can fix this later.

dagwieers:
@pietje666 In order to migrate from Travis CI to GitHub CI we need to disable Travis CI in the repository settings.

dagwieers:
There is another issue with using GitHub secrets. They do not seem to get picked up properly.

michaelarnauts:
@dagwieers you need to specify at every line in your step if you want to pass the variables, you have now commented them out?

dagwieers:
No, the credentials are only needed by the unit tests. And they are commented out because the tests fail with the credentials. So they are detected by the unit tests, but they  fail to work. I tried 3 sets of credentials.

dagwieers:
@michaelarnauts BTW I made some modifications to my workflow last night that you may like:

- Bring the Makefile in line with standards in VTM GO add-on
- Separate the different tasks in GitHub CI workflow
- Define environment variables at the job-level
- Disable fail-fast strategy

I decided to use the commands directly in the workflow, rather than call the make targets (except for translations, which is a bit more complex). Since some of them call coverage this also keeps the Makefile cleaner IMO.

Still not sure if I want to make the checks and the tests different jobs. Could be cleaner, but is less efficient for GitHub CI. May speed up the run time.

dagwieers:
So lets merge this and see what happens :-D

dagwieers:
Damn, forgot to remove my personal branch name before merging...

PR #705, opened 2020-03-25
Remove certain files from ZIP download

dagwieers:
This removes unneeded files from exported ZIP files.


michaelarnauts:
Ow, I didn't know this exists. That's really nice!

PR #708, opened 2020-03-25
Allow coverage steps to fail

dagwieers:
This is to ensure that our required pipelines are not blocked because of
Codecov or Sonarcloud availability issues.


dagwieers:
This is another case where we can only test in the master branch because of the need for a SONAR_TOKEN secret in GitHub CI :disappointed:

PR #713, opened 2020-03-26
Use SonarCloud action and Kodi addon-checker fixes

dagwieers:
This uses the newer options in the proposed pull-request.

### DIFF
--- a/.github/workflows/addon-check.yml
+++ b/.github/workflows/addon-check.yml
@@ -1,11 +1,7 @@
 name: Kodi
 on:
-  push:
-    branches:
-    - master
-  pull_request:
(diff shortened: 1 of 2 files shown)


dagwieers:
Maybe best to merge this only when SonarSource/sonarcloud-github-action#9 is merged.
I need to update the action with the upstream action at that time.

dagwieers:
I am going to merge this now as it helps a lot with the reliability of our tests.
We can switch back to upstream when things are merged over there.
- https://github.com/xbmc/addon-check/pull/217
- https://github.com/SonarSource/sonarcloud-github-action/pull/9

PR #714, opened 2020-03-26
Add credentials to service testing

dagwieers:
Because Michaël was right after all.


dagwieers:
Without this `favorites.is_activated()` would be false and the service would not be working.

PR #716, opened 2020-03-27
Add status workflow for VRT NU

dagwieers:
This adds a new workflow to test the availability of the VRT NU services
as well as the reliability of our own code during frequent testing..


dagwieers:
The past 30 minutes to an hour I am getting **HTTP 401 Unauthorized** errors on Python 3.5 only.
This is something we need to understand better...

PR #728, opened 2020-04-17
Check for unused translations

michaelarnauts:
I was just playing around with some things that we could check:

Current `master` branch has these unused translations:
```
No usage found for translation:
msgctxt "#30409"
msgid "episode"
msgstr ""

No usage found for translation:
msgctxt "#30957"
msgid ""
"Please increase the maximum bandwidth. Maximum bandwidth is set to {max} "
"kbps, but this stream needs a minimum of {min} kbps."
msgstr ""
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,7 @@ check-translations:
 	@-$(foreach lang,$(languages), \
 		msgcmp resources/language/resource.language.$(lang)/strings.po resources/language/resource.language.en_gb/strings.po; \
 	)
+	@tests/check-for-unused-translations.py
 
 check-addon: clean
 	@echo -e "$(white)=$(blue) Starting sanity addon tests$(reset)"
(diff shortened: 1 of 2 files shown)


michaelarnauts:
Note that this does an extremely naive check. I hope we never calculate with translation id's :)

dagwieers:
I wonder if there is a standardized framework for checks in Python?

dagwieers:
@michaelarnauts We prefer not to create branches in the main project, as they may get copied to other contributors and are not being cleaned up automatically. So it is best to create branches in your personal fork, and then make a PR from your personal fork branch to the master branch in this project.

(In fact we disallow this, but since you are an owner of Add-ons, you trump any restrictions, I think)

michaelarnauts:
Ow. Sorry. I'll make sure I setup my origins correctly.

dagwieers:
I think a check like this is definitely useful (also in the wider community), but I would like to see if we can fit this better in a framework rather than have it as a standalone script.

Also in some cases keeping unused translations might be deliberate and possibly temporary. That said, making this transparent is a win, always.

PR #729, opened 2020-04-26
Add basic M3U8 playlist file

dagwieers:
This is basic support for PVR IPTV Simple support in Kodi. It is a poor
man's implementation that (as it is) only adds VRT support (both TV and
Radio) to Kodi.

A better alternative is to use IPTV Manager in the future, nevertheless
this may be useful to some already as-is.


mediaminister:
Looks good to me, but I don't understand why VRT NU needs audio channels.

dagwieers:
Well, it helps a little bit because people can only add one single M3U8 to PVR IPTV Simple, so I took the liberty to add the VRT Radio channels as well. We can do this because they don't require an add-on.

mediaminister:
Okay, I understand now.

PR #732, opened 2020-04-27
Fix new pylint check

dagwieers:


mediaminister:
Kodi calls this `existingWindowId` and initiates it with a `-1` value https://codedocs.xyz/AlwinEsch/kodi/class_x_b_m_c_addon_1_1xbmcgui_1_1_window.html

PR #737, opened 2020-05-03
Add support for IPTV Manager

### LINKED ISSUE #209: Add TV livestreams to an IPTV PVR client
So the basic idea is that our users have their TV channels added to the TV section in their Kodi and can more easily switch between Live TV channels like they are used to on a normal TV.

To make this possible, various things should be considered:
- Either improve the existing IPTV Simple Client, or create a new IPTV PVR addon for this purpose
- The IPTV PVR addon needs an API to add/remove/query added streams
- Discuss with other stakeholders (other addon writers) for integration
- Either allow the user to do this on individual channels (right-click menu on channel) or do it automatically for all live channels
  - May cause duplicate entries if multiple overlapping addons, e.g. VRT NU vs Retrospect
- Optionally add EPG information when adding the channel
- Register the owner-addon of the channel
  - So that the user can right-click back to the addon for VOD content
- Automatically update entries when the VRT NU addon has changes (idempotent interface)

This solves the basic problem most users have that the TV channels are hidden inside addons, or need to be added to Favourites (where they are listed in between other stuff).

dagwieers:
This adds full IPTV Manager support including channel listings _and_ EPG data.
The EPG data is currently limited to yesterday, today and tomorrow.

This finally fixes #209


dagwieers:
@mediaminister I would like to merge this now so we can get more outside testing. This PR should not impact users, the only downside is that if you do not have IPTV Manager installed, the settings will have an entry in **VRT NU » Settings » Integration** to install it, which does not work (yet).

resources/lib/data.py
```diff
@@ -151,6 +160,7 @@
         youtube=[
             dict(label='Studio Brussel', url='https://www.youtube.com/user/StuBru'),
         ],
+        epg_logo='https://images.vrt.be/orig/2019/03/12/1e383cf5-44a7-11e9-abcc-02b7b76bf47f.png',
```
mediaminister, line 163: You can also use `height100`: https://images.vrt.be/height100/2019/03/12/1e383cf5-44a7-11e9-abcc-02b7b76bf47f.png

mediaminister:
It's okay to merge.

resources/lib/data.py
dagwieers: Thanks! I just decided to go with higher quality logos. And while I can find **width160**, **width320**, **width640** and **width1280**, these are often incorrectly cropped. So using the **orig** logos looks best.

PR #740, opened 2020-05-05
Add query string to URI in log_access

dagwieers:
This was always intended, but missing from the implementation. The
log_access() interface changes with this


dagwieers:
This is required for testing the IPTV Manager integration.

PR #743, opened 2020-05-08
Add epg_id

dagwieers:
Add a separate epg_id and synchronize this with VRT Radio EPG for the
live Radio channels.


dagwieers:
This is based on the channel id's and presets from: https://github.com/add-ons/service.iptv.manager/wiki/Channel-ids-and-presets

PR #751, opened 2020-05-18
Prepare for Python v3.9

dagwieers:
This is a test to see if Python v3.9 support is ready in GitHub CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
 #        max-parallel: 2
-        python-version: [ 2.7, 3.5, 3.6, 3.7, 3.8 ]
+        python-version: [ 2.7, 3.5, 3.6, 3.7, 3.8, 3.9 ]
     steps:
     - name: Check out ${{ github.sha }} from repository ${{ github.repository }}
       uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


dagwieers:
Starting with Python v3.10 the backward-compatibility layer with Python v2.7 will be removed. At that point it will be unlikely we can support both Python v2.7 and Python v3 in a single add-on.

Python v3.9 has not been released and Python v3.10 is expected roughly 18 months after the release of Python v3.9, so more likely we have 20 to 22 months before v3.10 becomes a thing.

michaelarnauts:
> Starting with Python v3.10 the backward-compatibility layer with Python v2.7 will be removed. At that point it will be unlikely we can support both Python v2.7 and Python v3 in a single add-on.
> 
> Python v3.9 has not been released and Python v3.10 is expected roughly 18 months after the release of Python v3.9, so more likely we have 20 to 22 months before v3.10 becomes a thing.

I can imaging that it will still be possible with tools like `six`, but I think it will be easier to just drop support for Python 2, or maybe create a branch for Python 2, and only do bugfixes there.

dagwieers:
> I can imaging that it will still be possible with tools like six, but I think it will be easier to just drop support for Python 2, or maybe create a branch for Python 2, and only do bugfixes there.

What we will do will largely depend on the user base at that time and the options we have when Python 3.10 is generally available.

PR #760, opened 2020-05-23
Add support for playing from the Guide with IPTV Manager

michaelarnauts:
This PR adds support to play programs from the PVR Guide through IPTV Manager.

Needs https://github.com/add-ons/service.iptv.manager/pull/19

### DIFF
--- a/resources/lib/iptvmanager.py
+++ b/resources/lib/iptvmanager.py
@@ -4,8 +4,9 @@
 """Implementation of IPTVManager class"""
 
 from __future__ import absolute_import, division, unicode_literals
-from kodiutils import log
+
 from data import CHANNELS
+from kodiutils import log
 
 
 class IPTVManager:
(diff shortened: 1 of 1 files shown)


michaelarnauts:
I don't know of an elegant way to not pass the `vod` field if the channel has no guide.

mediaminister:
With the dictionary update method?
```python



    @via_socket
    def send_channels():  # pylint: disable=no-method-argument
        """Return JSON-M3U formatted information to IPTV Manager"""
        streams = []
        for channel in CHANNELS:
            if not channel.get('live_stream_id'):
                continue
            stream = dict(
                id=channel.get('epg_id'),
                name=channel.get('label'),
                logo=channel.get('logo'),
                preset=channel.get('preset'),
                stream='plugin://plugin.video.vrt.nu/play/id/{live_stream_id}'.format(**channel),
            )
            if channel.get('has_tvguide'):
                stream.update(dict(vod='plugin://plugin.video.vrt.nu/play/airdate/{name}/{{date}}'.format(**channel)))

            streams.append(stream)
        return dict(version=1, streams=streams)
```

michaelarnauts:
> With the dictionary update method?

I was thinking in the lines as this, but I don't consider this an improvement actually.

```python
    @via_socket
    def send_channels():  # pylint: disable=no-method-argument
        """Return JSON-M3U formatted information to IPTV Manager"""
        streams = []
        for channel in CHANNELS:
            if not channel.get('live_stream_id'):
                continue
            streams.append(dict(
                id=channel.get('epg_id'),
                name=channel.get('label'),
                logo=channel.get('logo'),
                preset=channel.get('preset'),
                stream='plugin://plugin.video.vrt.nu/play/id/{live_stream_id}'.format(**channel),
            ).update(dict(
                vod='plugin://plugin.video.vrt.nu/play/airdate/{name}/{{date}}'.format(**channel),
            ) if channel.get('has_tvguide') else dict()))
        return dict(version=1, streams=streams)
```

I guess it could also be done with something like `dict(**dict(id=...,name=...), **(dict(vod=...) if channel.get('has_tvguide') else dict())` but I'll stick with the normal readable way.

dagwieers:
In fact, the `dict.update()` method accepts kwargs !

So this will work as well:
```python
            if channel.get('has_tvguide'):
                item.update(vod='plugin://plugin.video.vrt.nu/play/airdate/{name}/{{date}}'.format(**channel))
```

dagwieers:
Or we could allow a null value to indicate there is no VOD, next to not having the vod-key.
I think that would be a good practice anyway.

michaelarnauts:
The field is optional, and null values are interpreted as no vod, it was mainly just esthetics of not sending an empty vod line.

dagwieers:
Let's merge this. We can fix or improve later.

PR #761, opened 2020-05-27
This fixes an error on missing XSRF-TOKEN cookie

dagwieers:
I also cleaned up the flow to fail fast as fast returns make it easier
to understand the code without having to read the whole function.

This fixes the status errors as seen at: https://github.com/add-ons/plugin.video.vrt.nu/runs/712068916?check_suite_focus=true


dagwieers:
My "fix" now causes a more explicit error to the user:
```
ERROR: [plugin.video.vrt.nu] Failed to get usertoken from VRT NU
```

Issue #765, opened 2020-05-27
Selecting a "leaf" item within a VRT Nu menu yields an error.


tebokial:
### To Reproduce ###
1. Go to e.g. Most Recent
2. Click on Most Recent
3. VRT Nu Error 

### Additional context ###

- Operating system: Raspbian Stretch
- Kodi version: 18.1
- Addon version:  2.3.4
- Using a VPN: no
- Country you are using the addon from: Belgiekske

### Log (if available) ###
```paste below
[kodi.log](https://github.com/add-ons/plugin.video.vrt.nu/files/4689912/kodi.log)
2020-05-27 17:25:09.053 T:1596764912   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                             - NOTE: IGNORING THIS CAN LEAD TO MEMORY LEAKS!
                                            Error Type: <type 'exceptions.AttributeError'>
                                            Error Contents: 'NoneType' object has no attribute 'value'
                                            Traceback (most recent call last):
                                              File "/home/herman/.kodi/addons/plugin.video.vrt.nu/resources/lib/addon_entry.py", line 14, in <module>
                                                run(argv)
                                              File "/home/herman/.kodi/addons/plugin.video.vrt.nu/resources/lib/addon.py", line 336, in run
                                                plugin.run(argv)
                                              File "/home/herman/.kodi/addons/script.module.routing/lib/routing.py", line 130, in run
                                                self._dispatch(self.path)
                                              File "/home/herman/.kodi/addons/script.module.routing/lib/routing.py", line 141, in _dispatch
                                                view_func(**kwargs)
                                              File "/home/herman/.kodi/addons/plugin.video.vrt.nu/resources/lib/addon.py", line 197, in recent
                                                VRTPlayer().show_recent_menu(page=page)
                                              File "/home/herman/.kodi/addons/plugin.video.vrt.nu/resources/lib/vrtplayer.py", line 263, in show_recent_menu
                                                self._favorites.refresh(ttl=ttl('direct' if use_favorites else 'indirect'))
                                              File "/home/herman/.kodi/addons/plugin.video.vrt.nu/resources/lib/favorites.py", line 41, in refresh
                                                xvrttoken = TokenResolver().get_xvrttoken(token_variant='user')
                                              File "/home/herman/.kodi/addons/plugin.video.vrt.nu/resources/lib/tokenresolver.py", line 81, in get_xvrttoken
[...]

PR #767, opened 2020-05-28
Add a simple option for users to update VRT NU add-on

dagwieers:
This offers an easier path for users to check for updates or re-install
the VRT NU add-on.

This relates to #765

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -506,6 +506,14 @@ msgctxt "#30440"
 msgid "A-Z"
 msgstr ""
 
+msgctxt "#30450"
+msgid "Update add-on repositories"
+msgstr ""
+
+msgctxt "#30451"
+msgid "If a VRT NU add-on update is available, it will now be updated and you will be notified.\nMake sure the VRT NU add-on has [I]Auto-update[/I] enabled."
+msgstr ""
+
 
 ### SETTINGS
 msgctxt "#30700"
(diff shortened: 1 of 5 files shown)


resources/language/resource.language.en_gb/strings.po
```diff
@@ -506,6 +506,14 @@ msgctxt "#30440"
 msgid "A-Z"
 msgstr ""
 
+msgctxt "#30450"
+msgid "Update add-on repositories"
+msgstr ""
+
+msgctxt "#30451"
+msgid "If a VRT NU add-on update is available, it will now be updated and you will be notified.\nMake sure the VRT NU add-on has [I]Auto-update[/I] enabled."
```
mediaminister, line 514: What do you mean with "Make sure the VRT NU add-on has Auto-update enabled."?
Isn't this a global Kodi setting?

dagwieers: No, every add-on has a switch to enable/disable auto-update. We have seen that if you install an add-on from a ZIP file (that wasn't updated) this switch is off. But in most cases it will be on.

The problem is that in the OK window I only have 4 lines, so I cannot add a lot of information otherwise not everything is visible. Quite annoying.

mediaminister: I have disabled the global auto-update setting and then this individual add-on switch is not displayed. I think it's better to change the text to:

"Make sure you have _Auto-update_ enabled in Kodi and the VRT NU add-on."

dagwieers: Good suggestion! Having an **Updates** category for just one item is not very good, but burying it among other unrelated options in **Expert** is worse IMO. So I will do it like this for now, we can improve on this later.

dagwieers: BTW we can detect if the global setting is disabled, and we could offer to enable it (although I am not sure if this would be accepted by Team Kodi. We need to check. At least we can inform the user if this is a problem and maybe open the settings.

This is a future improvement.

PR #772, opened 2020-06-02
Add livestream cache

mediaminister:
This includes:
- Add livestream cache for faster zapping

### DIFF
--- a/resources/lib/kodiutils.py
+++ b/resources/lib/kodiutils.py
@@ -1184,6 +1184,13 @@ def get_url_json(url, cache=None, headers=None, data=None, fail=None, raise_erro
     return fail
 
 
+def generate_expiration_date(hours=2):
+    """Return ISO 8601 formatted expirationDate"""
+    from datetime import datetime, timedelta
+    import dateutil.tz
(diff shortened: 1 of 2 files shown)


michaelarnauts:
How much improvement does this do? I'm thinking to also cache the livestream tokens and URL to improve the playback of live TV on VTM Go, but from a quick look, it seems the most delay is due to the buffering in IA.

Maybe it's not worth the effort?

mediaminister:
It saves at least one http request after the user has clicked. I tested this and this can decrease the zapping delay with 490 ms.
Considering the zapping experience users had in the analog TV era, every 100 milliseconds you can gain is a noticeable improvement.

michaelarnauts:
> It saves at least one http request after the user has clicked. I tested this and this can decrease the zapping delay with 490 ms.
> Considering the instant zapping experience users had in the analog TV era, every 100 milliseconds you can gain is a noticeable improvement.

That's actually quite good! 👍

mediaminister:
Yes, http response time is still quite slow, this is documented: https://www.littledata.io/average/server-response-time

PR #777, opened 2020-06-10
Improve airdate api

mediaminister:
When start and end timestamps are provided, looking for the best matching "middle timestamp" to find an episode is more failsafe than just looking for the best matching start and end timestamps.

### DIFF
--- a/resources/lib/iptvmanager.py
+++ b/resources/lib/iptvmanager.py
@@ -48,7 +48,7 @@ class IPTVManager:
                 stream='plugin://plugin.video.vrt.nu/play/id/{live_stream_id}'.format(**channel),
             )
             if channel.get('has_tvguide'):
-                item.update(dict(vod='plugin://plugin.video.vrt.nu/play/airdate/{name}/{{date}}'.format(**channel)))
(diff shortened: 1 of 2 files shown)


resources/lib/iptvmanager.py
```diff
@@ -48,7 +48,7 @@ def send_channels():  # pylint: disable=no-method-argument
                 stream='plugin://plugin.video.vrt.nu/play/id/{live_stream_id}'.format(**channel),
             )
             if channel.get('has_tvguide'):
-                item.update(dict(vod='plugin://plugin.video.vrt.nu/play/airdate/{name}/{{date}}'.format(**channel)))
+                item.update(dict(vod='plugin://plugin.video.vrt.nu/play/airdate/{name}/{{start}}/{{stop}}'.format(**channel)))
```
michaelarnauts, line 51: Can we change this to `{{start}}/{{duration}}` and calculate the endpoint here? I agree that `date` isn't the best name for this field.

mediaminister: Currently, the VRT NU airdate interface uses two ISO 8601 timestamps. I would rather not change the current interface unless you have a very good reason.
In addition, duration is in seconds and I think mixing various time formats in one url is confusing.

michaelarnauts: This is fine, I can support both.

PR #778, opened 2020-06-10
Fix categories

mediaminister:
VRT NU added "Levensbeschouwing" category, changed html  and changed the name of the "Kinderen" category.
The status check didn't detect these changes because the fallback to hardcoded categories was wrongly placed inside the webscraper function.

This pull request fixes this and adds all new changes to the hardcoded categories.


dagwieers:
All good! Thanks for finding and fixing.

PR #781, opened 2020-06-15
Improve playing from TV guide

mediaminister:
This includes improvements to play episodes from the VRT NU add-on TV guide and IPTV TV guide:
- play using whatson_id and add metadata to the play item (instead of VRT NU url webscraping)
- fallback to airdate 24-hour livestream cache if there is no VOD stream available

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -526,6 +526,9 @@ msgctxt "#30453"
 msgid "Kodi is configured to not update add-ons automatically, so you will not get VRT NU updates unless you fix this setting.\nWe will open de Kodi settings, please change this setting."
 msgstr ""
 
+msgctxt "#30454"
+msgid "from livestream cache"
+msgstr ""
 
 
 ### SETTINGS
@@ -662,7 +665,7 @@ msgid "Streaming and DRM"
 msgstr ""
 
 msgctxt "#30787"
-msgid "Use Widevine DRM [COLOR=gray]for protected content)[/COLOR]"
+msgid "Use Widevine DRM [COLOR=gray](for protected content)[/COLOR]"
 msgstr ""
 
 msgctxt "#30788"
--- a/resources/lib/metadata.py
+++ b/resources/lib/metadata.py
@@ -739,27 +739,12 @@ class Metadata:
 
         # VRT NU Schedule API (some are missing vrt.whatson-id)
         elif api_data.get('vrt.whatson-id') or api_data.get('startTime'):
-            from datetime import datetime
-            import dateutil.parser
-            import dateutil.tz
             title = html_to_kodi(api_data.get('subtitle', '') or api_data.get('shortDescription', ''))
             label = '{start} - [B]{program}[/B]{title}'.format(
                 start=api_data.get('start'),
                 program=api_data.get('title'),
                 title=' - ' + title if title else '',
             )
-            now = datetime.now(dateutil.tz.tzlocal())
-            start_date = dateutil.parser.parse(api_data.get('startTime'))
-            end_date = dateutil.parser.parse(api_data.get('endTime'))
-            if api_data.get('url'):
-                if start_date <= now <= end_date:  # Now playing
-                    label = '[COLOR={highlighted}]%s[/COLOR] %s' % (label, localize(30301))
-            else:
-                # This is a non-actionable item
-                if start_date < now <= end_date:  # Now playing
-                    label = '[COLOR={greyedout}]%s[/COLOR] %s' % (label, localize(30301))
-                else:
-                    label = '[COLOR={greyedout}]%s[/COLOR]' % label
 
         # Not Found
         else:
--- a/resources/lib/tvguide.py
+++ b/resources/lib/tvguide.py
@@ -177,17 +177,26 @@ class TVGuide:
         episode_items = []
         for episode in episodes:
             program = url_to_program(episode.get('url', ''))
-            if episode.get('url'):
-                video_url = add_https_proto(episode.get('url'))
-                path = url_for('play_url', video_url=video_url)
-                context_menu, favorite_marker, watchlater_marker = self._metadata.get_context_menu(episode, program, cache_file)
-                label = self._metadata.get_label(episode) + favorite_marker + watchlater_marker
+            context_menu, favorite_marker, watchlater_marker = self._metadata.get_context_menu(episode, program, cache_file)
+            label = self._metadata.get_label(episode)
+            path = self.get_episode_path(episode, channel)
+            # Playable item
+            if '/play/' in path:
                 is_playable = True
+                label += favorite_marker + watchlater_marker
+            # Non-actionable item
             else:
-                label = '[COLOR={greyedout}]%s[/COLOR]' % self._metadata.get_label(episode)
-                path = url_for('noop', whatsonid=episode.get('vrt.whatson-id', ''))
-                context_menu, _, _ = self._metadata.get_context_menu(episode, program, cache_file)
                 is_playable = False
+                label = '[COLOR={greyedout}]%s[/COLOR]' % label
+
+            # Now playing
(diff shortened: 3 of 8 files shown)


resources/language/resource.language.en_gb/strings.po
```diff
@@ -662,7 +665,7 @@ msgid "Streaming and DRM"
 msgstr ""
 
 msgctxt "#30787"
-msgid "Use Widevine DRM [COLOR=gray]for protected content)[/COLOR]"
+msgid "Use Widevine DRM [COLOR=gray](for protected content)[/COLOR]"
```
dagwieers, line 668: I had noticed this as well, thanks !

resources/lib/metadata.py
```diff
@@ -222,11 +222,11 @@ def get_tvshowtitle(api_data):
 
         # VRT NU Search API
         if api_data.get('type') == 'episode':
-            return api_data.get('program', '???')
+            return api_data.get('program', '')
 
         # VRT NU Suggest API
         if api_data.get('type') == 'program':
-            return api_data.get('title', '???')
+            return api_data.get('title', '')
```
dagwieers, line 229: This was done intentionally, so that in case of incorrect metadata, entries would show something and it gives a point of reference to discuss and fix the cause.

Granted, it may not be the best placeholder, we could come up with something better, but an empty string is IMO a step back.

dagwieers:
With all recent changes under the hood, it feels like the next release ought to be a v2.4.0 release. I have lost track of any ramifications :wink:

resources/lib/metadata.py
mediaminister: It's just that I don't like triple question marks in Kodi GUI.

mediaminister: So, you think we need a placeholder in the GUI?

dagwieers: Oh I agree that triple question marks is not the best choice, but at least it is identifiable (on screen and in code).

Yes, if we ever get missing keywords where we do not expect them, we would like people to report this, rather than silently ignore. We can only do that by making this visually apparent. The problem by replacing it with "No description available", or "No title available" is that it might be seen as the actual content.

mediaminister: Okay, I'll remove this again.

mediaminister:
Okay, let's skip v2.3.6 and make v2.4.0 the next release than.

mediaminister:
Just fixed a bug. IPTV Manager expects `None` if there is no vod stream and VRT NU add-on expects noop plugin:// urls.

dagwieers:
@mediaminister Good point, we added noop-URLs only because we wanted to have non-actionable list-items and there was no easy way to do this. I think this is something we should actually take up with Kodi (and so this probably deserves an issue ticket).

Maybe there is a better way to achieve this already. I recently made the noop links unique so clicking on one would not jump the cursor to the first entry with the same URL. (This I missed when testing, I guess I always tested the first entry :roll_eyes:)

cc @michaelarnauts

michaelarnauts:
I've also noticed that for future episodes, there is no VOD stream provided. This causes that you can't play programs that were added within your IPTV Manager refresh timeframe.

I'm not sure if this is handled in this PR, or can be solved somehow.

mediaminister:
> I've also noticed that for future episodes, there is no VOD stream provided. 

There is a stream for episodes that have an url defined in the EPG data. It should work for episodes that show a play button at https://www.vrt.be/vrtnu/tv-gids/

> I'm not sure if this is handled in this PR, or can be solved somehow.

If you can provide an example of an VOD episode that is already available at VRT NU and doesn't show a play button in the VRT NU EPG, that would help to determine if this PR can be improved.

michaelarnauts:
I don't know any example of an program, but I think we could fallback to an airdate-url for future programs instead of providing None. I'm not sure if IPTV Manager should do this, or that the add-on can provide a airdate URL...

Do we know if a program will end up available as vod eventually but it is not yet at the moment of refreshing?

michaelarnauts:
I've noticed that the play button on the VRT Nu website can point to a previous episode (example Factcheckers at 10.45 on the 17th June, it points to an old episode), do we handle this to not use this URL then?

This is something that also happened on the EPG of Vier.

mediaminister:
> I think we could fallback to an airdate-url for future programs instead of providing None. I'm not sure if IPTV Manager should do this, or that the add-on can provide a airdate URL...

The airdate function will also fail for future programs because it uses the same VRT NU EPG data to determine if a program is available.

> Do we know if a program will end up available as vod eventually but it is not yet at the moment of refreshing?

Seems impossible. VRT API data is updated at the moment of publishing a new episode. Sometimes VRT only owns live broadcast rights and doesn't own VOD rights...

> I've noticed that the play button on the VRT Nu website can point to a previous episode (example Factcheckers at 10.45 on the 17th June, it points to an old episode), do we handle this to not use this URL then?

No, the whatson_id is used instead of the url, but the result is the same: this episode is playable

mediaminister:
Okay, I understand that airdate can help a little, but it will also disappoint users who expect a green bullet to indicate a playable program.

michaelarnauts:
That's right. Currently with the VTM GO addon, I don't even know if a program is playable before you try it, so you get a green bullet for everything. I don't know what is the right solution here...

It's also a strange user experience the last few hours are always unplayable.

mediaminister:
> I don't know what is the right solution here...

Realtime updating green dots?

michaelarnauts:
Those green dots are written to the xmltv file (in Leia, in Matrix they appear automatically because there is a direct vod url), do the only solution there is to refresh more often.

resources/lib/tvguide.py
```diff
@@ -203,6 +212,19 @@ def get_episode_items(self, date, channel):
...
 
+    @staticmethod
+    def get_episode_path(episode, channel, noop=False):
+        """Return a playable plugin:// path for an episode"""
+        now = datetime.now(dateutil.tz.tzlocal())
+        end_date = dateutil.parser.parse(episode.get('endTime'))
+        if episode.get('url') and episode.get('vrt.whatson-id'):
+            return url_for('play_whatson_id', whatson_id=episode.get('vrt.whatson-id'))
+        if now - timedelta(hours=24) <= end_date <= now:
+            return url_for('play_air_date', channel, episode.get('startTime')[:19], episode.get('endTime')[:19])
```
michaelarnauts, line 223: I don't think we want this. We are not sure that the play by airdate will work, and we will fallback to the airdate url in IPTV Manager anyway.

When we supply an url, we will also see a green bullet in the EPG, and this should be used to confirm that there is a program available.

dagwieers:
So after a discussion with @michaelarnauts to understand all situations with the existing add-ons. We don't know what will work, so we cannot indicate what will work. We only know three things:
- We have a direct link
- We have an indirect link (airdate) for the channel
- We have no link (usually there is no EPG either)

We cannot really take the future into account because these indicators (in Leia) are pre-processed. So I would go for this:
- A direct link → green dot (more probable it will work)
- An indirect link → yellow dot (less probable it will work)
- No link → red dot or no dot?

The reason why I would always show a dot (even if we have an airdate) is because users may think no dot means it will not work, but that really depends on the add-on. It could be helpful if an add-on could give an indication to the probability that airdate URLs work, so we could differentiate with yellow and maybe orange, but on the other hand this might get confusing quickly.

Fact is that despite the colours, we have no guarantees what will work so we need to make this very clear. Maybe by showing an info-window on first startup (and a settings item that also show this window on demand). And info in the README.

resources/lib/tvguide.py
mediaminister: Okay, I'll remove this.

mediaminister:
I changed the PR so IPTV Manager can decide how and when to use the airdate api. VRT NU only returns valid VOD urls or `None`

michaelarnauts:
Perfect! I think this PR is fine to merge! We should continue the discussion regarding the bullets in a new issue on IPTV Manager.

PR #785, opened 2020-06-18
Fix season items

mediaminister:
This fixes https://github.com/add-ons/plugin.video.vrt.nu/issues/783

It turned out that `seasons` is not a valid `xbmcplugin.setContent` type and this caused problems with the Arctic Horizon skin


dagwieers:
Owh, I just noticed this was updated just minutes before. I hope this was ready?

mediaminister:
No problem, it was ready just minutes before.

PR #787, opened 2020-06-19
Add asserts to stubs

dagwieers:
It is probably a good idea to check for argument types and values where
this is possible to fail fast.

### DIFF
--- a/tests/xbmc.py
+++ b/tests/xbmc.py
@@ -14,6 +14,11 @@ import weakref
 from xbmcextra import ADDON_ID, global_settings, import_language
 from utils import to_unicode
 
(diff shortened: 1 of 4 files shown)


dagwieers:
@mediaminister @michaelarnauts I am sure we can add more of such checks to avoid common issues. If you have ideas, let me know.

I think these might help beginning add-on developers.

michaelarnauts:
Seems like a good idea although I was wondering if it might be a good time to start looking at https://github.com/retrospect-addon/kodi.emulator.ascii for testing.

It's a project of @basrieter and it is about the same as what we are doing with our `xbmc*` implementations.

PR #791, opened 2020-07-03
Return to directory listing when stream is None

mediaminister:
If there is no valid stream to play with Kodi Player, we should always return to the directory listing. Otherwise a busy dialog is shown forever.


dagwieers:
We currently get an error when playing the last episode of Terzake (from yesterday evening).
The stream could not be found: https://www.vrt.be/vrtnu/a-z/terzake/2020/terzake-d20200702/

Issue #281, opened 2019-05-29
Some on demand programs are not played until the end.


mediaminister:
When VRT serves VOD-programs from a livestream([Live to VOD](https://www.google.com/search?q=%22live+to+vod%22)) with the so called "[virtual subclips](http://docs.unified-streaming.com/documentation/vod/player-urls.html#virtual-subclips)", the VRT NU add-on doesn't play the stream till the end of the program. 

This bug occurs mostly with live current affair programs like "Het Journaal", "Terzake", "De Afspraak", and "De Zevende Dag" while VRT broadcasts them.

You can bypass this bug by waiting to watch the program until VRT has broadcasted the program completely.

dagwieers:
The experience I have had just now is that if an on-going live program is listed, it is often listed as a 02:00:00 (2-hour) video. This has 2 consequences I noticed:
- If you watch it 15 minutes after the start, it appears to stop after 15 minutes, if you then rewatch 45 minutes after the start, it stops after 45 minutes
- If the live stream overruns those 2 hours, it will stop at 2 hours and it may take some time before a longer running entry (in this case 02:25:05) appears with the full live stream

It is seldom a good idea to watch a live video delayed, if you watch only 15 minutes too late, you can start the live TV stream and rewind 15 minutes to watch the start without hiccups until the end.

dagwieers:
Discussing this on Facebook I wonder how VRT NU's web player handles this. If this is reproducible, we can report this upstream. If other players handle this better, we can report this upstream as a point of improvement (inputstream.adaptive).

mediaminister:
I'll check this in a couple of minutes: https://www.vrt.be/vrtnu/a-z/het-journaal/2020/het-journaal-het-journaal-13u-20200720/

mediaminister:
VRT NU's web player dynamically updates the total time while playing a Live-to-VOD stream URL if only a start timestamp is provided: https://live-cf-vrt.akamaized.net/groupc/live/8edf3bdf-7db3-41c3-a318-72cb7f82de66/live.isml/.mpd?t=2020-07-20T11:07:00

But it seems Kodi (IA) always needs a stop timestamp **in the past** to start a Live-to-VOD stream at the start timestamp. In all other cases Kodi(IA) treats the stream as an ordinary livestream.

mediaminister:
[...]

PR #794, opened 2020-07-20
Fix stop timestamp in virtual subclip

mediaminister:
This PR adds some clarification about virtual subclip support in Unified Origin streaming platform. It also ensures that a past stop timestamp is always added to the stream url when a virtual subclip is needed because Unified Origin streaming platform fails to return a virtual subclip with a missing or future stop timestamp.


dagwieers:
This relates to #281

PR #797, opened 2020-07-23
Fix unwanted virtual subclip behaviour

mediaminister:
This fixes https://github.com/add-ons/plugin.video.vrt.nu/issues/281

This should be tested thoroughly during live broadcasts of "Het Journaal" (preferably 10 minutes after the live broadcast has started using the "Most recent" menu)
- 13:00 CEST: Het journaal 13u 
- 19:00 CEST: Het journaal 19u
- Sometime after 22:00 CEST: Het journaal Laat

This fix only works with InputStream Adaptive (when Widevine DRM is enablad in the add-on settings)

### DIFF
--- a/resources/lib/playerinfo.py
+++ b/resources/lib/playerinfo.py
@@ -9,7 +9,7 @@ from xbmc import getInfoLabel, Player, PlayList
 from apihelper import ApiHelper
 from data import CHANNELS
 from favorites import Favorites
-from kodiutils import addon_id, get_setting_bool, has_addon, kodi_version_major, log, notify, set_property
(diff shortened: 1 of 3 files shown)


dagwieers:
I have not had a chance to test this (and the original problem) yet. But I still plan to do this.

mediaminister:
There's no hurry. I still have to test how this impacts HLS streams.
For now, it should work with MPEG-DASH. (Widevine DRM enabled)

mediaminister:
This PR won't benefit or hurt HLS streams which are directly played in Kodi to preserve Kodi 17 compatibility. Kodi Player doesn't detect the start timestamp.
This is ready to be merged as works well with MPEG-DASH and InputStream Adaptive.

PR #806, opened 2020-08-30
Fix exception on SSLError/log_error

dagwieers:
I experienced the following exception:

```
                                              File "/storage/.kodi/addons/plugin.video.vrt.nu/resources/lib/kodiutils.py", line 1128, in open_url
                                                log_error('{error}\nurl: {url}', error=exc.reason, url=url)
                                            AttributeError: 'SSLError' object has no attribute 'reason'
```

What is strange is that this is on a Python 2.7.16 (LibreELEC v9.2.8)
while according to the documentation this should not be happening.

I implemented 2 workarounds, based on possible values for 2.7.8 and
earlier releases.

### DIFF
--- a/resources/lib/kodiutils.py
+++ b/resources/lib/kodiutils.py
@@ -1123,9 +1123,19 @@ def open_url(url, data=None, headers=None, method=None, cookiejar=None, follow_r
         ok_dialog(heading=localize(30968), message=localize(30969))
         log_error('URLError: {error}\nurl: {url}', error=exc.reason, url=url)
         return None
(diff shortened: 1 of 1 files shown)


dagwieers:
I could not test this as the issue was resolved in the meantime.
But we check everything to ensure we use the proper values, so I think this is fine.

mediaminister:
LibreELEC v9.2.8? Does this exist?

dagwieers:
> LibreELEC v9.2.8? Does this exist?

A verifiable error, I meant v9.2.4 🙄

PR #809, opened 2020-08-31
Handle none-responses from open_url

dagwieers:
This handles all none-responses from open_url().

I am not sure this is the best way to handle this, it is not very
elegant. Raising an exception helps in one case, but would not be
elegant in the other cases either.

### DIFF
--- a/resources/lib/tokenresolver.py
+++ b/resources/lib/tokenresolver.py
@@ -199,7 +199,10 @@ class TokenResolver:
         )
         data = dumps(payload).encode()
         headers = {'Content-Type': 'application/json', 'Cookie': login_cookie}
-        setcookie_header = open_url(self._TOKEN_GATEWAY_URL, data=data, headers=headers).info().get('Set-Cookie')
+        response = open_url(self._TOKEN_GATEWAY_URL, data=data, headers=headers)
+        if response is None:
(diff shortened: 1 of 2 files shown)


resources/lib/tokenresolver.py
```diff
@@ -213,13 +216,22 @@ def _get_roaming_xvrttoken(self):
...
+        response = open_url(req_info.get('Location'), follow_redirects=False)
+        if response is None:
+            return None
+        url = response.info().get('Location')
         headers = {'Cookie': cookie_value}
         if url is None:
             return None
-        setcookie_header = open_url(url, headers=headers, follow_redirects=False).info().get('Set-Cookie')
+        response = open_url(url, headers=headers, follow_redirects=False)
+        if response ise None:
```
michaelarnauts, line 232: Typo here

dagwieers: Thanks for linting, but I also got the GitHub CI mail ;-)

dagwieers:
So I will merge this. My concern is that `open_url()` may return None without raising an issue. But as far as I can see this should never happen, so I am fine with this, even if it is not very elegant, I don't see a better solution to exit a calling function prematurely on network issues.

PR #814, opened 2020-09-18
Get categories from online json

mediaminister:
While experimenting with **mitmproxy** I found out that the VRT NU Android app uses json files to display categories and channels.
This PR deprecates web scraping for the categories menu and uses https://www.vrt.be/vrtnu/categorieen/jcr:content/par/categories.model.json as a source instead.


dagwieers:
Wh00t w00t !

michaelarnauts:
This is really nice!

PR #815, opened 2020-09-28
Add Kodi library support

dagwieers:
This adds support for adding VRT NU as a Kodi library source.

As a result Movies and TV shows from VRT NU will be listed in the Kodi Movies and TV shows sections.


dagwieers:
After a discussion with @michaelarnauts, there is also an option to leave up to the user what Movies and TV shows are integrated, by adding:
- plugin://plugin.video.vrt.nu/library/my-movies
- plugin://plugin.video.vrt.nu/library/my-tvshows

Then the user can decide whether they want to include only the Favorites, or instead all the Movies and TV shows. So this is something we have to decide, do we have a flag in the settings to change the behaviour, or a different target?

mediaminister (changes requested):
I only get favourite tv-shows with `plugin://plugin.video.vrt.nu/library/tvshows`

> Then the user can decide whether they want to include only the Favorites, or instead all the Movies and TV shows. So this is something we have to decide, do we have a flag in the settings to change the behaviour, or a different target?

I think it is confusing to use a setting that changes the behaviour of the plugin url. Using different plugin urls is better. Especially when these plugin urls are used completely outside the add-on environment.

> This adds support for adding VRT NU as a Kodi library source.

I'm not familiar with this functionality with regard to add-ons, does this have advantages over using the add-on itself?

dagwieers:
> I only get favourite tv-shows with `plugin://plugin.video.vrt.nu/library/tvshows`

That is intentional

> > Then the user can decide whether they want to include only the Favorites, or instead all the Movies and TV shows. So this is something we have to decide, do we have a flag in the settings to change the behaviour, or a different target?
> 
> I think it is confusing to use a setting that changes the behaviour of the plugin url. Using different plugin urls is better. Especially when these plugin urls are used outside the add-on environment.
> 
> > This adds support for adding VRT NU as a Kodi library source.
> 
> I'm not familiar with this functionality with regard to add-ons, does this have advantages over using the add-on itself?

michaelarnauts:
I think the most clear experience would be to have the same behaviour for Movies and TV Shows and export only the items in your favourites by default. It seems like an odd choice to do it by default for series, but not for movies

dagwieers:
> I think the most clear experience would be to have the same behaviour for Movies and TV Shows and export only the items in your favourites by default. It seems like an odd choice to do it by default for series, but not for movies

Well, on VRT NU a movie is usually only around for 7 days, so most people would never "Follow" or "Watch Later" a movie. In most cases if they were not aware, they would miss it. The nice things about the Kodi Library is that these new Movies (we only have a handful) would appear in the "Recently added " list.

Same for TV Shows, but there having all episodes listed would be a poor signal-to-noise ratio.

So for Movies, if you don't add all, you have no signal and no noise, but for TV shows adding them all, would mean mostly noise.

dagwieers:
Thinking about Netflix, I would probably do the same thing. I would like to know when Netflix adds new Movies, but I am only interested in new episodes of the stuff I am watching.

dagwieers:
I just noticed 5 new movies were released by Streamz:
- Assassination Island
- Pachamama
- Ruin Me
- True Grit
- Waves

Our kids will only see **Pachamama** on their Movies screen.

dagwieers:
I would like to get this ready for an upcoming v2.5.0 release, based on the implementation and settings from VTM GO and Streamz.

deurenkletser:
I tried adding the sources to Kodi, but I'm always getting an error: 'Gedeeld op afstand: pad niet gevonden of ongeldig'.
Is it possible to add the sources so tv series appear in my library?

mediaminister:
Kodi library support is not yet fully implemented, so this won't work.
I have currently no spare time to implement this.
Feel free to rewrite this code to get a working implementation.

deurenkletser:
Thank you for clearing this out. Unfortunately have no code writing skills, but thanks for keeping the plug-in updated!

PR #823, opened 2020-10-20
Fix integrated landscape view in Arctic Horizon skin

mediaminister:
This fixes https://github.com/add-ons/plugin.video.vrt.nu/issues/822 and reintroduces https://github.com/add-ons/plugin.video.vrt.nu/issues/783

Using `xbmcplugin.setContent('seasons')` enables the Integrated Landscape View, but in that case [Arctic Horizon skin](https://github.com/jurialmunkey/skin.arctic.horizon) doesn't pick up the `ListItem` metadata in Kodi Player, so this PR introduces https://github.com/add-ons/plugin.video.vrt.nu/issues/783 again.

### DIFF
--- a/resources/lib/apihelper.py
+++ b/resources/lib/apihelper.py
@@ -162,7 +162,7 @@ class ApiHelper:
         season_items = []
         sort = 'label'
         ascending = True
-        content = 'files'
+        content = 'seasons'
 
         episode = random.choice(episodes)
         program_type = episode.get('programType')
@@ -668,7 +668,7 @@ class ApiHelper:
                 label = channel.get('label')
                 plot = '[B]%s[/B]' % channel.get('label')
                 is_playable = False
-                info_dict = dict(title=label, plot=plot, studio=channel.get('studio'), mediatype='video')
+                info_dict = dict(title=label, plot=plot, studio=channel.get('studio'))
                 stream_dict = []
                 prop_dict = {}
             elif channel.get('live_stream') or channel.get('live_stream_id'):
@@ -691,7 +691,7 @@ class ApiHelper:
                 else:
                     plot = localize(30142, **channel)  # Watch live
                 # NOTE: Playcount and resumetime are required to not have live streams as "Watched" and resumed
-                info_dict = dict(title=label, plot=plot, studio=channel.get('studio'), mediatype='video', playcount=0, duration=0)
+                info_dict = dict(title=label, plot=plot, studio=channel.get('studio'), playcount=0, duration=0)
                 prop_dict = dict(resumetime=0)
                 stream_dict = dict(duration=0)
                 context_menu.append((
@@ -744,7 +744,7 @@ class ApiHelper:
                     label = '[B]%s[/B]' % label
                 plot = localize(30144, **youtube)  # Watch on YouTube
                 # NOTE: Playcount is required to not have live streams as "Watched"
-                info_dict = dict(title=label, plot=plot, studio=channel.get('studio'), mediatype='video', playcount=0)
+                info_dict = dict(title=label, plot=plot, studio=channel.get('studio'), playcount=0)
 
                 context_menu = [(
                     localize(30413),  # Refresh menu
@@ -773,7 +773,7 @@ class ApiHelper:
                 label=featured_name,
                 path=url_for('featured', feature=feature.get('id')),
                 art_dict=dict(thumb='DefaultCountry.png'),
-                info_dict=dict(plot='[B]%s[/B]' % feature.get('name'), studio='VRT'),
+                info_dict=dict(title=feature.get('name'), plot='[B]%s[/B]' % feature.get('name'), studio='VRT'),
             ))
         return featured_items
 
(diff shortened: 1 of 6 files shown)


resources/lib/apihelper.py
```diff
@@ -162,7 +162,7 @@ def __map_seasons(self, program, seasons, episodes):
         season_items = []
         sort = 'label'
         ascending = True
-        content = 'files'
+        content = 'seasons'
```
dagwieers, line 165: We did this IIRC to ensure folders were shown as folders, and not as files.
Which in fact is a general Kodi issue.

resources/lib/vrtplayer.py
```diff
@@ -34,49 +34,49 @@ def show_main_menu(self):
...
+                      info_dict=dict(title=localize(30026), plot=localize(30027))),
             TitleItem(label=localize(30028),  # Search
                       path=url_for('search'),
-                      art_dict=dict(thumb='DefaultAddonsSearch.png'),
-                      info_dict=dict(plot=localize(30029))),
+                      art_dict=dict(thumb='DefaultAddonsSearch.png', fanart='DefaultAddonsSearch.png'),
+                      info_dict=dict(title=localize(30028), plot=localize(30029))),
         ])
-        show_listing(main_items, cache=False)  # No category
+        show_listing(main_items, cache=False, content='tvshows')  # No category
```
dagwieers, line 79: This will show "TV shows / VRT NU" instead of "Videos / VRT NU" in the breadcrumbs of the main menu

resources/lib/apihelper.py
mediaminister: If you don't set the mediatype on the ListItem, you get a folder icon in the Kodi skin.

dagwieers: I know a lot of add-ons doing this wrong then :-D

mediaminister: I don't understand the point of `mediatype` either.

resources/lib/vrtplayer.py
mediaminister: I changed it to 'videos'.

resources/lib/metadata.py
```diff
@@ -547,12 +547,13 @@ def get_year(api_data):
...
     def get_mediatype(self, api_data, season=False):
-        """Get art dict from single item json api data"""
+        """Get mediatype from single item json api data"""
 
         # VRT NU Search API
         if api_data.get('type') == 'episode':
+            # Don't set mediatype on seasons to get a folder icon in the Kodi skin
             if season:
-                return 'season'
+                return ''
```
michaelarnauts, line 556: How important is that folder icon? I think it would make sense to set the content type correctly on what it is, The way how it's rendered in the default skin is probably a design choice?

michaelarnauts: This is not a remark, just a question. I think I'm also doing stuff like this in VTM GO.

mediaminister: It's probably a bug in the Estuary-skin. The content type is set with `xbmcplugin.setContent()` I'm still figuring out what the purpose of the `mediatype` infoLabel is.

I think it's important that the user experience with the default skin is consistent.

mediaminister:
@vermope @piejanssens Can you test this PR again please with the most recent https://github.com/jurialmunkey/skin.arctic.horizon

I think https://github.com/add-ons/plugin.video.vrt.nu/issues/783 is fixed. If you can confirm, I think this PR can be merged.

vermope:
@mediaminister Works perfectly! Seems it's the default view now, did not had to change the view. Thanks!!

mediaminister:
I got the same https://github.com/add-ons/plugin.video.vrt.nu/issues/783 bug again. It's not fixed apparently, but the bug only occurs when you play an episode after selecting the season in the integrated landscape view.

dagwieers (approved):
This  is looking good to me.

dagwieers:
We may want to verify if there is a bug in Estuary wrt. folder icons.

piejanssens:
Tested the build from the zip provided by @mediaminister in this PR and I don't see a difference between Arcitc Horizon master and 0.5.58.

Both showing this OSD while playing VOD:
Schermafbeelding 2021-02-11 om 08 01 08

piejanssens:
My bad, I knew it was too early 😴 
I can report that the issue is still as described in #583

mediaminister:
I finally found the cause of https://github.com/add-ons/plugin.video.vrt.nu/issues/783
The missing OSD info is related to the use of the built-in `PlayMedia` function in https://github.com/jurialmunkey/skin.arctic.horizon/blob/master/1080i/Includes_View_53_Seasons.xml
When using `PlayMedia`, there is no metadata passed from the selected `ListItem`. To get this fixed, the add-on needs to request the metadata again:

> I found out skin.video.siren is setting metadata in the play function to workaround this bug. This kind of workaround is not desirable in the VRT NU add-on because we need an extra API request for that and this will slow down player start.

https://github.com/add-ons/plugin.video.vrt.nu/issues/822#issuecomment-717771303

PR #831, opened 2020-11-05
Support xbmcvfs.translatePath

mediaminister:
In Kodi 19 alpha 2 `xbmcvfs.translatePath` was added and this will replace `xbmc.translatePath` in future versions of Kodi.

### DIFF
--- a/resources/lib/kodiutils.py
+++ b/resources/lib/kodiutils.py
@@ -11,6 +11,12 @@ from ssl import SSLError
 import xbmc
 import xbmcaddon
 import xbmcplugin
+
+try:  # Kodi 19 alpha 2 and higher
+    from xbmcvfs import translatePath
+except ImportError:  # Kodi 19 alpha 1 and lower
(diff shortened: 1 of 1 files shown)


dagwieers:
I would prefer this implementation, which is import-bound and not runtime-bound.

https://github.com/emilsvennesson/script.module.inputstreamhelper/commit/bef0bff26bd7fc733a62ba77d3ea6261a120bba8

mediaminister:
Okay, I changed it and added an extra check becasue we don't need unicode conversion at all in Kodi 19.

dagwieers:
> Okay, I changed it and added an extra check because we don't need unicode conversion at all in Kodi 19.

I don't agree 100% because the unicode conversion is linked to the python version and not the Kodi version. We can assume that Kodi v19 is always Python 3, but that check is already in to_unicode, so we shouldn't have another check for Kodi v19 IMO.

mediaminister:
Indeed, that extra check is not really necessary. I removed it.

PR #836, opened 2020-12-13
Support programs with more than 10 seasons

mediaminister:
This fixes https://github.com/add-ons/plugin.video.vrt.nu/issues/835

### DIFF
--- a/resources/lib/apihelper.py
+++ b/resources/lib/apihelper.py
@@ -601,6 +601,19 @@ class ApiHelper:
             facets = search_json.get('facets', {}).get('facets')
             if facets:
                 seasons = next((f.get('buckets', []) for f in facets if f.get('name') == 'seasons' and len(f.get('buckets', [])) > 1), None)
+            # Experimental: VRT Search API only returns a maximum of 10 seasons, to get all seasons we need to use the "model.json" API
+            if seasons and program and len(seasons) == 10:
+                season_json = get_url_json('https://www.vrt.be/vrtnu/a-z/%s.model.json' % program)
+                season_items = None
+                try:
+                    season_items = season_json.get(':items').get('parsys').get(':items').get('container') \
+                                              .get(':items').get('banner').get(':items').get('navigation').get(':items')
+                except AttributeError:
+                    pass
+                if season_items:
+                    seasons = []
+                    for item in season_items:
+                        seasons.append(dict(key=item.lstrip('0')))
 
         episodes = search_json.get('results', [{}])
         show_seasons = bool(season != 'allseasons')


michaelarnauts:
Can't we just always use the `model.json` API?

mediaminister:
That is a possibility but it requires research. I don't know yet if the `model.json` API is some kind of standard API. This will consume time which I don't have right now.

resources/lib/apihelper.py
```diff
@@ -601,6 +601,14 @@ def get_episodes(self, program=None, season=None, episodes=None, category=None,
             facets = search_json.get('facets', {}).get('facets')
             if facets:
                 seasons = next((f.get('buckets', []) for f in facets if f.get('name') == 'seasons' and len(f.get('buckets', [])) > 1), None)
+            # Experimental: VRT Search API only returns a maximum of 10 seasons, to get all seasons we need to use the "model.json" API
+            if seasons and program and len(seasons) == 10:
+                seasons = []
+                season_json = get_url_json('https://www.vrt.be/vrtnu/a-z/%s.model.json' % program)
+                season_items = season_json.get(':items').get('parsys').get(':items').get('container') \
+                                          .get(':items').get('banner').get(':items').get('navigation').get(':items')
```
dagwieers, line 609: My concern is that this can easily break. It feels as if the data structure was modelled for a very specific use in the front-end, and for that reason not a good API to be used for anything structural.

In this case I am not so much concerned with that it stops working as it is a very marginal issue and a workaround, but more about how it will show an error that the user may not understand or feel the add-on is not reliable. So I would prefer either that it continues as before without throwing an error, or gives a pop-up that is clear the user should report it.

dagwieers:
Ideally we would have the back-end return more than 10 seasons. It is not unlikely that VRT NU added this limit arbitrarily and are now affected by it in some way they have not realised themselves. So I would not be surprised that this limit is suddenly lifted. (Which is fine, the implementation would only cause issues if it is exactly 10)

resources/lib/apihelper.py
mediaminister: Thanks, I added a try...except block to catch the AtrributeError.
When the json would change, it continues as before.

mediaminister:
This "model.json" API is also used in the VRT NU Android app to get the seasons. I did a quick search and found out it is an API from Adobe Experience Manager CMS: https://experienceleague.adobe.com/docs/experience-manager-65/developing/components/json-exporter.html#components

dagwieers:
> This "model.json" API is also used in the VRT NU Android app to get the seasons. I did a quick search and found out it is an API from Adobe Experience Manager CMS: https://experienceleague.adobe.com/docs/experience-manager-65/developing/components/json-exporter.html#components

Ok, so it is not going to change as easily as I expected. That is good. Still it does not seem a good idea to use this as a basis for anything else. Using more than one API backend increases the probability that the add-on may be impacted. For a workaround this is ok.

In this case, I think it could be worthwhile to also report this to VRT NU to explain the issue and look for a fix. We could test if this limitation is removed as well so we know when we can get rid of the workaround in the future.

dagwieers (approved):
I am happy with this, the impact on the code is limited (always good for a workaround) and it is well-documented. So it should stay out of the way mostly, and if there are issues there is a trace to why it is there. Well done !

mediaminister:
> In this case, I think it could be worthwhile to also report this to VRT NU to explain the issue and look for a fix. We could test if this limitation is removed as well so we know when we can get rid of the workaround in the future.

I'm not sure this is a VRT NU API bug. I guess we can get all seasons from the API if we request all pages. Currently, only the first page with 300 episodes is requested to determine the seasons.

PR #837, opened 2020-12-17
Fixes a date parsing issue on windows and run CI

michaelarnauts:
* Run CI on Windows (Python 2.7, Python 3.8 and Python 3.9)

* Fixes `datelong` formatting that currently is `%A, %e %B %Y`, but `%e` isn't a standard in Python. Under certain situations (like the setup of CI), this doesn't work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,26 +9,41 @@ on:
 jobs:
   tests:
     name: Add-on testing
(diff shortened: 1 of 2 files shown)


michaelarnauts:
This should be okay now. The `%e` is now manually parsed, since it's not supported on all platforms.

Apologies for the messy history, but I didn't have a windows environment to test on so I had to use Github :)

PR #847, opened 2021-01-11
Use InputStream Adaptive to play HLS

mediaminister:
Play all streams using InputStream Adaptive from Kodi 18 onwards. This fixes https://github.com/add-ons/plugin.video.vrt.nu/issues/846

To test this, you need to disable "Use Widevine DRM".


dagwieers:
I had to rebase the branch to be able to test this as v2.4.2 with the other changes since v2.4.0.

IIRC previously we could influence MPEG-DASH or HLS by switching ISA on and off.
How would we be able to do that now?

mediaminister:
> IIRC previously we could influence MPEG-DASH or HLS by switching ISA on and off.
> How would we be able to do that now?

No, you're right, "Use InputStream Adaptive" in Expert settings still works to disable IA.

Issue #854, opened 2021-01-26
timeshift functionality 


siemon-geeroms:
Hi, 

Maybe it's a technical limitation but why is Kodi not allowing/able to pause a live stream?
or is it a setting that I'm overlooking?

thanks in advanced,

Siemon

dagwieers:
It works fine here. Can you describe a bit more what you are doing?

siemon-geeroms:
When watching the één live stream via de VRT nu add-on on kodi 18 (libreelec) I only have the option to stop the stream but not pause it.

I will have a look at that setting you mentioned later this evening.

siemon-geeroms:
Ok the InputStream Adaptive add-on was not even installed. after installing and enabling it in the add-on I can pause live streams. Thanks for your help!

dagwieers:
Thanks for reporting, I didn't even know pause was not working on live streams without ISA.

We now have a record of it 😉

PR #855, opened 2021-01-26
Inform user that no time shift is possible without ISA

siemon-geeroms:


siemon-geeroms:
related to issue #854

dagwieers:
@siemon-geeroms We don't want to require **Inputstream Adaptive** as it is not strictly necessary, and **InputStream Helper** should take care of everything for the end-user. There might be a problem with distributions that have Kodi installed but are missing the additional **inputstream.adaptive** add-on.

siemon-geeroms:
I understand the hesitation to make it mandatory but now the user doesn't get a hint that this is the cause of not being able to timeshift. 

Suggestion: display a one time message first time the addon is openend informing the user of this? Also maybe add it to the wiki in the installation section

michaelarnauts:
Hmm, didn't we get at a point where we want to require it for Leia+, but don't for Krypton?

dagwieers:
> Hmm, didn't we get at a point where we want to require it for Leia+, but don't for Krypton?

I don't think we want to have 3 distributions for this. When we drop Krypton support, we can reconsider.

That said, this is a very marginal issue. Most Kodi users do have **inputstream.adaptive** simply because all other (Flemish) add-ons basically require it for installation. And if you don't, pause/timeshift fails only for live streams.

PS I cannot check our usage statistics anymore, but last time I checked we had about 10% Krypton users IIRC. Which was 10x more than Matrix then.

dagwieers (changes requested):
The proposed change will break Krypton and older Leia installations.

Personally I don't think the issue is important enough to warrant fixing like this. Most users are not affected as they need ISA for other add-ons. We could opt to show a pop-up on start-up as suggested, one time only (on upgrade/install once). We have a framework for this already.

siemon-geeroms:
I wanna have a crack at it because it's a good entry point to start dev on kodi addons. But can you point me to the framework you mention for displaying one time only messages? Because the only approach I see, is to add a hidden setting that will be checked/set if the message is already displayed and ISA is not installed on start up off the addon. 

> The proposed change will break Krypton and older Leia installations.
> 
> Personally I don't think the issue is important enough to warrant fixing like this. Most users are not affected as they need ISA for other add-ons. We could opt to show a pop-up on start-up as suggested, one time only (on upgrade/install once). We have a framework for this already.

mediaminister (commented):
VRT NU provides multiple stream formats. One of them is HLS (with or without AES encryption) and this works directly in Kodi Player. We decided that we want to support this in case the user hasn't InputStream Adaptive installed or doesn't want to use the proprietary Widevine DRM decryption module.
This is also needed for Kodi 17 Krypton support.

So, adding InputStream Adaptive as a requirement is not the right solution.

I guess a warning on first run can help, the function is already there: https://github.com/add-ons/plugin.video.vrt.nu/blob/27fd834b49ff42c761fc2e0fccd6a77ccd2a988e/resources/lib/vrtplayer.py#L98-L116

Another approach is adapting streamservice.py and the play function in kodiutils.py to make better use of [InputStream Helper](https://github.com/emilsvennesson/script.module.inputstreamhelper) which can also show a warning or automatically install InputStream Adaptive. But the impact on different systems needs to be thoroughly tested as we still want to support systems that don't use InputStream Adaptive.

resources/language/resource.language.en_gb/strings.po
```diff
@@ -1042,3 +1042,7 @@ msgstr ""
 msgctxt "#30987"
 msgid "No stream found, please try again!"
 msgstr ""
+
+msgctxt "#30988"
+msgid "InputStream Adaptive not installed/enabled, pause/play and timeshift of live streams will not work"
```
mediaminister, line 1047: Please add a verb: InputStream Adaptive **is**

resources/lib/vrtplayer.py
```diff
@@ -93,6 +93,10 @@ def _version_check(self):
                 # 2.2.1 version: moved tokens: delete old tokens
                 from tokenresolver import TokenResolver
                 TokenResolver().delete_tokens()
+            
+            # Make user aware that timeshift funxtionalty will not work without ISA when user starts up the first time
```
mediaminister: proper spelling is functionality

mediaminister (dismissed):

resources/language/resource.language.en_gb/strings.po
```diff
@@ -1042,3 +1042,7 @@ msgstr ""
 msgctxt "#30987"
 msgid "No stream found, please try again!"
 msgstr ""
+
+msgctxt "#30988"
+msgid "InputStream Adaptive is not installed/enabled, pause/play and timeshift of live streams will not work"
```
dagwieers, line 1047: The phrasing confuses me because of the punctuation. I would change this into:

```suggestion
msgid "InputStream Adaptive is not available, therefore pause and timeshift of live streams will not work"
```

If ISA is not enabled, InputStream Helper will take care of it. Also, playing live streams will work so the statement was factual incorrect IMO.

resources/lib/vrtplayer.py
```diff
@@ -9,7 +9,7 @@
 from kodiutils import (colour, delete_cached_thumbnail, end_of_directory, get_addon_info,
                        get_setting, get_setting_bool, get_setting_int, has_credentials,
                        localize, log_error, ok_dialog, play, set_setting, show_listing,
-                       ttl, url_for, wait_for_resumepoints)
+                       ttl, url_for, wait_for_resumepoints, has_inputstream_adaptive, kodi_version_major)
```
dagwieers, line 12: We tend to sort the imports as it is easier to find an entry in the list.

dagwieers (commented):
Looks good, minor suggestions only.

resources/language/resource.language.en_gb/strings.po
siemon-geeroms: I agree for the most part but prefer to use "installed" instead of "available" because this gives the idea that it is missing and can't be installed.
```suggestion
msgid "InputStream Adaptive is not installed, therefore pause and timeshift of live streams may not work"
```

resources/language/resource.language.nl_nl/strings.po
```diff
@@ -1042,3 +1042,7 @@ msgstr "Geen on demand stream beschikbaar voor {title}."
 msgctxt "#30987"
 msgid "No stream found, please try again!"
 msgstr "Geen stream gevonden, gelieve opnieuw te proberen!"
+
+msgctxt "#30988"
+msgid "InputStream Adaptive is not installed/enabled, pause/play and timeshift of live streams will not work"
```
dagwieers, line 1047: We need to update this as well.
```suggestion
msgid "InputStream Adaptive is not installed, therefore pause and timeshift of live streams may not work"
```

resources/language/resource.language.nl_nl/strings.po
```diff
@@ -1042,3 +1042,7 @@ msgstr "Geen on demand stream beschikbaar voor {title}."
...
+msgstr "InputStream Adaptive is niet geinstalleerd/actief, pauzeren/afspelen en timeshift van live tv zullen niet werken"
```
dagwieers, line 1048:
```suggestion
msgstr "InputStream Adaptive is niet geïnstalleerd, daarom kan het pauzeren en timeshift van live tv niet werken"
```

dagwieers (changes requested):
Update and translation.

resources/language/resource.language.nl_nl/strings.po
dagwieers: This is not resolved, this is the English `msgid` in the Dutch translation. If this is not identical, it will not be translated to Dutch.

dagwieers (approved):
All good for me! Will need to test this night.

mediaminister (approved):

PR #873, opened 2021-03-29
Use own kodi-addon-submitter

mediaminister:


dagwieers:
Since this is something that impacts all our add-ons. Maybe we should host it on the add-ons project?

mediaminister:
This does nothing yet and can be reversed. I still have to find the cause of a bug in kodi-addon-submitter.

PR #878, opened 2021-04-21
Add Tubed integration

mediaminister:


dagwieers:
Looks good. I have not been able to test it myself (no Kodi v19 atm).

mediaminister:
It doesn't work yet. I'm not sure if `plugin://plugin.video.tubed/user/KetnetVideo/` is the right url scheme. 
And I get the message "quota exceeded".

PR #893, opened 2021-07-20
Enable Widevine DRM by default

mediaminister:
It seems VRT removed the HLS AES livestreams, this means you need Widevine CDM module to play livestreams. Setting Widevine DRM setting enabled by default makes sure the user can play livestreams without touching the settings.


dagwieers:
This is likely because VRT now has 1080p for live streams and some licenses require it for HD content.

PR #981, opened 2022-05-12
Fix channels menu

mediaminister:


ghen2:
These latest commits fix the recent issues for me, thanks!

jekkos:
These last commits indeed fix the broken plugin. 
I had to package and install manually. A release would be nice to have the plugin pushed through the update channel again.
Thanks for this.

mediaminister:
I'll release next week.

PR #997, opened 2022-06-02
Trim program queries to 24 digits

mediaminister:


mediaminister:
This fixes https://github.com/add-ons/plugin.video.vrt.nu/issues/996

PR #1012, opened 2022-08-23
VRT MAX rebranding

mediaminister:


resources/lib/apihelper.py
```diff
@@ -500,7 +500,7 @@ def get_latest_episode(self, program):
         api_data = self.get_episodes(program=program, variety='single')
         if len(api_data) != 1:
             return None
-        episode = api_data[0]
```
michaelarnauts, line 503: This seems to be the only non-branding change here. Is this intentional? Maybe you want to merge this earlier untill VRT Max is actually live?

mediaminister: Thanks for noticing.

Issue #1209, opened 2025-02-27
Not able to browse programs - even after upgrading to latest version


PeterCloet:
### Describe the bug ###
Issue browsing the programs - even after upgrading to 2.5.35 (latest version)

### Log (if available) ###
https://paste.kodi.tv/aziducobip.kodi

Addon version: 2.5.35
Kodi version: 20.3
Inputstream adaptive version: v20.3.18.1
Operating System (Windows / Mac OS / Android / LibreElec / OSMC / ...): LibreElec 11.0.6
Special Hardware (RPI / Vero4K+ / ...): RPI

PeterCloet:
@mediaminister I see you made the changes for the latest release. Can you have another look?

yorickps:
Can confirm, exactly the same issue...

malc0mn:
```txt
2025-03-01 20:26:05.100 T:136800   error <general>: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                                    - NOTE: IGNORING THIS CAN LEAD TO MEMORY LEAKS!
                                                   Error Type: <class 'routing.RoutingError'>
                                                   Error Contents: No route to path "/programs"
                                                   Traceback (most recent call last):
                                                     File "/Users/<user>/Library/Application Support/Kodi/addons/plugin.video.vrt.nu/resources/lib/addon_entry.py", line 14, in <module>
                                                       run(argv)
                                                     File "/Users/<user>/Library/Application Support/Kodi/addons/plugin.video.vrt.nu/resources/lib/addon.py", line 326, in run
                                                       plugin.run(argv)
                                                     File "/Users/<user>/Library/Application Support/Kodi/addons/script.module.routing/lib/routing.py", line 130, in run
                                                       self._dispatch(self.path)
                                                     File "/Users/<user>/Library/Application Support/Kodi/addons/script.module.routing/lib/routing.py", line 143, in _dispatch
                                                       raise RoutingError('No route to path "%s"' % path)
                                                   routing.RoutingError: No route to path "/programs"
                                                   -->End of Python script error report<--
```

malc0mn:
Added PR that will fix this issue.

mediaminister:
Fixed and released in https://github.com/add-ons/plugin.video.vrt.nu/releases/tag/v2.5.36

PR #1210, opened 2025-03-01
Update utils.py to support additional vrtmax cases

malc0mn:
Fix for issue #1209.

### DIFF
--- a/resources/lib/utils.py
+++ b/resources/lib/utils.py
@@ -100,26 +100,6 @@ def reformat_image_url(url):
     return ''
 
 
-def program_to_url(program, url_type):
-    """Convert a program url component (e.g. de-campus-cup) to:
-        - a short programUrl (e.g. /vrtnu/a-z/de-campus-cup/)
-        - a medium programUrl (e.g. //www.vrt.be/vrtnu/a-z/de-campus-cup/)
-        - a long programUrl (e.g. https://www.vrt.be/vrtnu/a-z/de-campus-cup/)
-   """
-    url = None
-    if program:
-        # short programUrl
-        if url_type == 'short':
-            url = '/vrtnu/a-z/' + program + '/'
-        # medium programUrl
-        elif url_type == 'medium':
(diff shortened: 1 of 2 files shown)


resources/lib/utils.py
malc0mn, line 1: The `program_to_url`function is only used in unit tests, so this should be dropped.

resources/lib/utils.py
```diff
@@ -100,26 +100,6 @@ def reformat_image_url(url):
     return ''
 
 
-def program_to_url(program, url_type):
```
malc0mn, line 103: This was entirely unused, so I removed it rather than support vrtmax urls as well.

resources/lib/utils.py
```diff
@@ -136,7 +116,7 @@ def url_to_program(url):
     elif url.startswith('//www.vrt.be/vrtnu/a-z/'):
         # medium programUrl or targetUrl
         program = url.split('/')[5]
-    elif url.startswith('/vrtnu/a-z/'):
+    elif url.startswith('/vrtnu/a-z/') or url.startswith('/vrtmax/a-z/'):
```
malc0mn, line 119: Used or instead of dropping 'vrtnu' since this would be a BC break for users with local HTTP cache.

resources/lib/utils.py
```diff
@@ -158,7 +138,7 @@ def url_to_episode(url):
     if url.startswith('//www.vrt.be/vrtnu/a-z/'):
         # medium episode url
         return url.replace('//www.vrt.be/vrtnu/a-z/', '/vrtnu/a-z/')
-    if url.startswith('/vrtnu/a-z/'):
+    if url.startswith('/vrtnu/a-z/') or url.startswith('/vrtmax/a-z/'):
```
malc0mn, line 141: Used or instead of dropping 'vrtnu' since this would be a BC break for users with local HTTP cache.

mediaminister:
Many thanks for this fix.

PR #1318, opened 2026-03-21
Decode season_name and update list_id construction

driesenj:
Decode season_name from base64 format and update list_id accordingly.


resources/lib/api.py
```diff
@@ -2316,6 +2316,7 @@ def get_recent_episodes(end_cursor='', use_favorites=False):
 
 def get_episodes(list_id=None, destination=None, end_cursor='', program_name=None, season_name=None, use_favorites=False,
                  feature=None, keywords=None, date=None, channel=None):
+    from base64 import b64decode
```
mogwa1: Minor nitpicking probably, but is this import still needed?
It doesn't seem to be referenced anymore.

PR #1334, opened 2026-05-18
fix resumepoints

Simon-D-R:
I switch between Kodi and browser sometimes so it's useful to be able to resume between one and the other.

It seems the extraction of the episode_id from the path was the main issue. Tested locally and confirmed working now.

(I also started fixing tests, but it seems some were failing on master, so I just committed the one I had fixed by then, no new test failures introduced)


mediaminister:
Thanks for this contribution. Your code seems fine and I'm going to merge this for now.
Some tests currently fail because VRT removed some metadata from their api.

